PR #1, opened 2022-02-03
chore: add GitHub action

bigint:
Screenshot 2022-02-03 at 6 08 03 PM

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,0 +1,23 @@
+name: Node CI
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        node-version: [14.x]
+
+    steps:
+      - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,23 @@
...
+
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+      - name: yarn install, build, lint
+        run: |
+          yarn
```
mktcode, line 21:
```suggestion
          yarn --frozen-lockfile
```

> If you need reproducible dependencies, which is usually the case with the continuous integration systems, you should pass --frozen-lockfile flag.
https://classic.yarnpkg.com/lang/en/docs/cli/install/

bigint: thanks, how i forgot it 🙈

mktcode (approved):

bigint:
cc @bonustrack

Issue #142, opened 2022-06-11
Send token modal for execution


bonustrack:
On the editor to create a proposal here:
https://goerli.snapshotx.xyz/#/0x0625dc1290b6e936be5f1a3e963cf629326b1f4dfd5a56738dea98e1ad31b7f3/create
We should do some changes in the send token modal for execution:

- To select a token, the modal should change to something like this (colors and size are a bit off you can use same what we use in the UI currently)
<img width="609" alt="image" src="https://user-images.githubusercontent.com/16245250/173205633-e605fbd9-747a-4953-8d7c-1990a43d196b.png">
When we click on a token it should comeback to the initial state of the modal with the token selected. We could have an icon to go back to initial state, we use https://heroicons.com for this, the icon can be "arrow-narrow-left".

- We should load the list of tokens based on the space treasury balances, you can find an example of query tokens in the Treasury.vue view using Covalent API.  

- We should make it in a way that when you type USD value it update the Amount and vice versa.

PR #152, opened 2022-06-18
feat: TokenPicker to SendToken modal

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/142

Changes:
- extract `useBalances` composable
- update `INumber` component as there is no way to make it update from the outside (modalValue) is just the default value and turned it in more standard two-way bound input that accepts `value` and `@input`. This makes it compatible with `v-model` (which is not used here as we need to control value updates).
- add `TokenPicker`
- make amount/value inputs adjust based on the rate.

## Test plan
- Go to http://localhost:8080/#/0x0625dc1290b6e936be5f1a3e963cf629326b1f4dfd5a56738dea98e1ad31b7f3/create
- Type some amount, nothing happens.
- Pick some token.
- Value input has updated.
- Try using both inputs, they should be linked.
- If you delete all contents  from one input it's also deleted from other (so placeholder shows up).

### DIFF
--- a/src/components/Block/TokenPicker.vue
+++ b/src/components/Block/TokenPicker.vue
@@ -5,6 +5,7 @@ import { _n } from '@/helpers/utils';
 
 const props = defineProps({
   searchValue: String,
+  loading: Boolean,
   assets: Array
 });
 
@@ -25,7 +26,10 @@ const filteredAssets = computed(() =>
 </script>
 
 <template>
-  <div v-if="filteredAssets.length === 0" class="text-center py-2">
+  <div v-if="loading" class="px-4 py-3 block flex justify-center">
+    <UiLoading />
+  </div>
+  <div v-if="!loading && filteredAssets.length === 0" class="text-center py-2">
     No results
   </div>
   <div
--- a/src/components/Block/TokenPicker.vue
+++ b/src/components/Block/TokenPicker.vue
@@ -29,30 +29,34 @@ const filteredAssets = computed(() =>
   <div v-if="loading" class="px-4 py-3 block flex justify-center">
     <UiLoading />
   </div>
-  <div v-if="!loading && filteredAssets.length === 0" class="text-center py-2">
-    No results
-  </div>
-  <div
-    v-for="(asset, i) in filteredAssets"
-    :key="i"
-    role="button"
-    class="px-3 py-1 border-b last:border-0 flex justify-between"
-    @click="emit('pick', asset.contract_address)"
-  >
-    <div class="flex items-center">
-      <Stamp type="token" :id="asset.contract_address" :size="32" />
-      <div class="flex flex-col ml-3 leading-[20px] text-skin-link text-sm">
-        <span>
-          {{ asset.contract_ticker_symbol }}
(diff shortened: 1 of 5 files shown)


src/components/Block/TokenPicker.vue
```diff
@@ -0,0 +1,54 @@
...
+import { _n } from '@/helpers/utils';
+
+const props = defineProps({
+  searchValue: String,
+  assets: Array
+});
+
+const emit = defineEmits(['pick']);
+
+const filteredAssets = computed(() =>
```
Sekhmet, line 13: We could use `fuse.js` to make it smarter (and sort results), but I kept is basic to avoid extra deps.

bonustrack: Interesting I didn't knew about Fuse.js

src/components/S/INumber.vue
```diff
@@ -1,23 +1,28 @@
...
 const props = defineProps({
-  modelValue: Number,
+  value: [Number, String],
   definition: Object
 });
 
-const emit = defineEmits(['update:modelValue']);
+const emit = defineEmits(['input']);
 
-const input = ref(props.modelValue || props.definition.default || undefined);
```
bonustrack, line 11: Any reason to change this file? This code is used on others place we should keep ability to define default in `props.definition.default`

bonustrack (commented):
Looks great! Can you do these changes too:
- Do not load the balances when the editor page open but only when you open token picker modal
- Add the token image with 27px size here: 
image
- Don't make this a text input but instead make it a link (this way text is not selected and we get the hand cursor).  <br/>
- Add a "Max" helper to populate amount input like this:
image
- Make the search input focus directly when you open the token picker modal

src/components/S/INumber.vue
Sekhmet: I changed it because:
> update INumber component as there is no way to make it update from the outside (modalValue) is just the default value and turned it in more standard two-way bound input that accepts value and @input. This makes it compatible with v-model (which is not used here as we need to control value updates).

But now I see that `IString` works the same way, but it works somehow, I guess I need to learn more about Vue to figure out how it works (and why it doesn't work there).

bonustrack: Ops sorry missed your explanations, I like your way better if it works and if we don't lose the ability to default to `props.definition.default`. Your way could be used in others components in S/ folder too. These components in S/ folder are made to be reusable on others repo to generate form based on JSONSchema.

bonustrack:
I'm able to get both the loading indicator and the list of tokens, it should be either one or another
image

Also the spacing in red should be the same as in green on the input here:
image

Sekhmet:
> Also the spacing in red should be the same as in green on the input here:

I increased it because the icon didn't fit with suggested size (27) - at least it doesn't look good to me.



We can keep it the same size, but we will need to make the icon smaller then.

bonustrack:
@Sekhmet You can use a smaller icon for now, I'll try to do some tweak in the design later

bonustrack (approved):
Works well!

Issue #153, opened 2022-06-18
Send NFT modal for execution


bonustrack:
Similar than send token modal we need a send NFT modal and selector. The select would look like this:
<img width="636" alt="image" src="https://user-images.githubusercontent.com/16245250/174451825-88895a76-feb6-45bc-88d8-0ccea09de2f7.png">
Showing squared images with the collection name and token Id. The image should also appear in the "Send NFT" modal. It's possible to get the list of all NFT that the account own using Covalent API (similar than tokens)

PR #161, opened 2022-06-19
feat: add NFT picker

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/153

Adds NFT picker to Send NFT modal.

**Note**: Sometimes Covalent is slow to fetch NFTs.

## Test plan

- Replace `loadNfts(spaces.pasta.wallets[0]);` with `loadNfts('0xfC00116B4049BBc8618847C74B44F0e103D571A3');` or other address you want to test with.
- You can pick, search and select NFTs.
- Preview shows up in the send NFT modal.

### DIFF
--- a/src/components/NftPreview.vue
+++ b/src/components/NftPreview.vue
@@ -5,7 +5,9 @@ const props = defineProps({
 });
 
 function handleError(e) {
-  e.target.src = `https://cdn.stamp.fyi/token/${props.item.contract_address}?s=256`;
+  const size = props.size ? props.size * 2 : 256;
+
+  e.target.src = `https://cdn.stamp.fyi/token/${props.item.contract_address}?s=${size}`;
 }
 </script>
 
--- a/src/components/NftPreview.vue
+++ b/src/components/NftPreview.vue
@@ -5,7 +5,7 @@ const props = defineProps({
 });
 
 function handleError(e) {
-  e.target.src = `https://stamp.fyi/token/${props.item.contract_address}?s=256`;
+  e.target.src = `https://cdn.stamp.fyi/token/${props.item.contract_address}?s=256`;
 }
 </script>
 
--- a/src/components/NftPreview.vue
+++ b/src/components/NftPreview.vue
@@ -1,0 +1,26 @@
+<script setup>
+const props = defineProps({
+  item: Object,
+  size: Number
+});
+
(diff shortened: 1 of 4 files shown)


src/composables/useNfts.ts
```diff
@@ -0,0 +1,46 @@
...
+        const title =
+          item.nft_data[0]?.external_data?.name ??
+          item.contract_name ??
+          'Untitled';
+        const displayTitle =
+          title.match(/(#[0-9]+)$/) || !id ? title : `${title} #${id}`;
+
+        const image =
+          item?.nft_data[0]?.external_data.image ??
+          `https://stamp.fyi/token/${item.contract_address}?s=256`;
```
Sekhmet, line 27: Would be nice to have Stamp support for NFTs.

bonustrack: Ok yes, feel free to add an issue for this, we could add support with url like this `stamp.fyi/nft/$address/$tokenId`. Also we should always use the cdn subdomain `cdn.stamp.fyi` to display image, at some point we will have a landing page at stamp.fyi. But currently I dont think we should use stamp here with the token url.

bonustrack: Unless you are using it to get Blockie fallback image?

Sekhmet: Here it's only to load the fallback (blockie) image in two cases:
1. image is missing on the NFT
2. image is there, but it fails to load ([here](https://github.com/snapshot-labs/sx-ui/blob/9809780a3f8a08381f8f10716e6bc169ee1abb28/src/components/NftPreview.vue#L7-L9)).

src/components/NftPreview.vue
```diff
@@ -0,0 +1,26 @@
+<script setup>
+const props = defineProps({
+  item: Object,
+  size: Number
+});
+
+function handleError(e) {
+  e.target.src = `https://cdn.stamp.fyi/token/${props.item.contract_address}?s=256`;
```
bonustrack, line 8: I'd recommend to use a 2x size here to get more precision and optimize to not load too big image
```suggestion
  e.target.src = `https://cdn.stamp.fyi/token/${props.item.contract_address}?s=${props.size * 2}`;
```

bonustrack (approved):
Looks good!

Issue #162, opened 2022-06-19
Add execution txs from send tokens and send NFT forms


bonustrack:
When someone create execution txs using send token or send NFT we should store the transaction in an `execution` array. The format of the array should look like this:
https://gist.github.com/bonustrack/5d4551421ac570d472c07450f503eb21

Here is some explaination about the different fields:
- "to" is the contract you interact with in the tx, most of the time its a ERC20 or NFT contract unless you send ETH in which case "to" is just the recipient.
- "data" this is the encoded arguments of function called in the smart contract, for example for doing a transfer the args will be the recipient and amount, this is encoded using this function <https://github.com/snapshot-labs/sx-ui/blob/db27536ad7ed659990d0c044ce2dc880f2e593a8/src/components/Modal/Transaction.vue#L29-L31> (It require the ABI of the contract function, for NFT and ERC20 it's standardized so you can use the same all the time). Here is for ERC20 transfer:
```const abi = ["function transfer(address _to, uint256 _value) public returns (bool success)"]```
For NFT ```const abi = ["function transferFrom(address _from, address _to, uint256 _tokenId) external payable"]``` (may need to confirm for NFT transfer)
- "value" is the amount of ETH that you are sending in the tx, when sending ERC20 or NFT it would be 0, but if you transfer ETH the amount would be stored in this "value" field.

The others fields are just metadata that we can use to repopulate the Send Token or Send NFT forms.

bonustrack:
When a tx is added it can be displayed like this in the execution section:
<img width="680" alt="image" src="https://user-images.githubusercontent.com/16245250/174486323-1aa6de1d-9ed9-4fef-8dc8-e519c0fc6bb5.png">

PR #163, opened 2022-06-20
feat: make SendToken/SendNft/Contract modals generate txs

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/162

Changes:
- Modified `INumber` component. `modelValue` is what it needed to be (just implementation was breaking two-way binding, now we use new implementation, with old prop names, compatible with `v-model`).
- Added `ERC20`, `ERC721`, `ERC1155` ABIs.
- Added `Transaction` type to represent intended format for transactions.
- Enabled TS support in `SendToken`, `SendNft` and `ExecutionEditable` so we can verify that it conforms to the type. Skipped `Transaction` for not as it looks like it needs more changes and I don't want to pollute this PR even more.
- Made `SendToken`, `SendNft` and `Transaction` output tx in the same format.
- Updated `ExecutionEditable` to display those transactions types properly.

## Test plan
- Add send token execution (with ERC20 and ETH).
- Add send NFT execution.
- Add contract call execution.
- All of them should look in execution list as expected.
- Inspect generated `tx` objects that contents look good.

## Comments

Submitting modals might fail (for example amount is invalid or contract call is invalid for some reason). It would be nice to display some kind of error if that happens. I left it out of this PR as it's pretty big already.


src/helpers/abis/erc1155.json
```diff
@@ -0,0 +1,314 @@
+[
```
bonustrack, line 1: We should add only ABI for the functions that we use, also we should use the string ABI it's much more compact, see this code to turn ABI to this compact mode https://github.com/snapshot-labs/snapshot.js/blob/master/test/abi.ts

src/components/Modal/SendToken.vue
```diff
@@ -70,6 +80,46 @@ function handleMaxClick() {
...
+
+  const isEth =
+    currentToken?.value.contract_address ===
+    '0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee';
+
+  let data = '0x';
+  if (!isEth) {
+    const iface = new Interface(erc20);
+    data = iface.encodeFunctionData(
+      'safeTransferFrom(address, address, uint256)',
```
bonustrack, line 97: This should use `transfer` method, `safeTransferFrom` is for NFTs
```suggestion
      'transfer(address, uint256)',
```

bonustrack (commented):
Added few suggestions, would be nice also to move some logic as we can into an helper file to keep the views simple

bonustrack (approved):
Works well!

Issue #167, opened 2022-06-21
Send token form improvement


bonustrack:
We should have Ethereum selected by default when the form is open, also when on the token picker ETH should appear first regardless of the $ value.

PR #171, opened 2022-06-21
feat: prioritize and preselect Ethereum in picker

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/167

- Ethereum will be shown first in the list of assets in the picker
- Ethereum will be selected by default

### DIFF
--- a/src/components/Modal/SendToken.vue
+++ b/src/components/Modal/SendToken.vue
@@ -3,6 +3,7 @@ import { reactive, ref, computed, watch, nextTick, Ref } from 'vue';
 import { formatUnits } from '@ethersproject/units';
 import { useBalances } from '@/composables/useBalances';
 import { createSendTokenTransaction } from '@/helpers/transactions';
+import { ETH_CONTRACT } from '@/helpers/constants';
 import { clone } from '@/helpers/utils';
 
 const props = defineProps({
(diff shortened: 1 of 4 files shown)


src/components/Modal/SendToken.vue
```diff
@@ -23,7 +24,20 @@ const showPicker = ref(false);
...
+  const token = assetsMap.value?.get(form.token);
+
+  if (!token) {
+    return {
+      contract_decimals: 18,
+      contract_name: 'Ether',
+      contract_ticker_symbol: 'ETH',
+      contract_address: ETH_CONTRACT
+    };
+  }
```
Sekhmet, line 37: Fallback to static token as we only load list of assets when user actually opens the picker.

bonustrack (approved):
Looks good!

src/components/Modal/SendToken.vue
Sekhmet: @bonustrack just noticed that because it's missing quote, to USD value input is not going to work unless you open token picker first.
Should we fetch all balances when modal loads to avoid this? I'm not sure how can we do that otherwise (unless we somehow only load ETH rate here to make inputs work as expected).

PR #177, opened 2022-06-23
Improve style for execution modals

bonustrack:
Mostly changing spacing for search input and list of tokens, also change "Select token" to "Select NFT" on the Send NFT modal.


Sekhmet (approved):
tACK

Issue #165, opened 2022-06-21
Make execution txs editable


bonustrack:
When adding a new tx (with send token, nft or contract call) and confirm, the tx will appear on the execution section. We should make each tx on this section editable. We can add an edit icon in the tx preview and when user click on it it will open the related modal (token or nft or contract) populated with the current tx parameters.

PR #180, opened 2022-06-25
feat: make proposal transactions editable

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/165

- Update `IObject` and `IString` to properly do two-way binding (same as changes in INumber before).
- Update TransactionModal to use `computed` values so it better handles data coming from two places.
- Bugfixes in TransactionModal that handle issues when switching methods on already populated forms or when ABI gets updated.
- Add edit option to all transaction types.

## Test plan
- Add bunch of transactions (token, nfts, contract calls).
- Try editing some of them (changing recipients, values, tokens, different contracts, methods etc.).
- Nothing should go terribly wrong.

I used those two contracts to get ABIs:
- `0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6`
- `0x11fE4B6AE13d2a6055C8D9cF65c55bac32B5d844`

## Screenshot

image

### DIFF
--- a/src/components/Block/ExecutionEditable.vue
+++ b/src/components/Block/ExecutionEditable.vue
@@ -8,6 +8,8 @@ import { Transaction } from '@/types';
 const address = spaces.pasta.wallets[0];
 
 const txs: Ref<Transaction[]> = ref([]);
+const editedTx: Ref<number | null> = ref(null);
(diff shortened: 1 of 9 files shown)


bonustrack (approved):
Works perfectly for me, well done!

Issue #166, opened 2022-06-21
Store execution on the proposal and draft


bonustrack:
Currently the execution is not added in the payload of the proposal metadata, we should add it and also have it stored on drafts

PR #181, opened 2022-06-26
feat: store proposal transactions

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/166

## Test plan
- Add some transactions to proposal.
- Refresh page, they are still there.
- Switch between different drafts, it restores transaction for given proposal.

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -42,10 +42,10 @@ export function useActions() {
     title: string,
     body: string,
     discussion: string,
-    transactions: Transaction[]
+    execution: Transaction[]
   ) {
     if (!web3.value.account) return await forceLogin();
-    const pinned = await pin({ title, body, discussion, transactions });
+    const pinned = await pin({ title, body, discussion, execution });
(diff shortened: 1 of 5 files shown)


bonustrack (commented):
For send NFT modal the form does not keep his state for me
image

src/composables/useActions.ts
```diff
@@ -40,10 +41,11 @@ export function useActions() {
     executionHash: string,
     title: string,
     body: string,
-    discussion: string
+    discussion: string,
+    transactions: Transaction[]
   ) {
     if (!web3.value.account) return await forceLogin();
-    const pinned = await pin({ title, body, discussion });
+    const pinned = await pin({ title, body, discussion, transactions });
```
bonustrack, line 48: Sorry just noticed but could we rename `transactions` to `execution` in draft and pinned file?
```suggestion
    const pinned = await pin({ title, body, discussion, execution });
```

bonustrack:
Now it works well with NFT too

bonustrack (approved):
tACK

Issue #164, opened 2022-06-20
Add validation on S/ input components


bonustrack:
We should check if an input is valid or not using JSON schema definition validation with AVJ library on the S/ input components. If there is an error the input should have a class "s-input-error" and text color would be red. On definition with "format": "address" we should check if the address is correct with a custom validation using @ethersproject/address isAddress function. We can also add custom validation for "uint256" format. The style of the error should not be defined in the component but instead on style.scss, components should stay style agnostic.

PR #208, opened 2022-07-13
feat: add form validation

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/164

Adds validation for forms using ajv schema.

- Validation is done at top level and components just react to `error` prop. It could be possible to do it at the component level (as those components already receive definitions), but getting validation status out of deeper components back seems weird and it seems most libraries do it at top level.
- Fixed components checking wrong prop for defaultValue, became apparent after enabling TS.

## Test plan
- Go to Contract call modal.
- Invalid contract address should give you error message.
- Input correct address.
- Inputting invalid address or uint256 into args should result in error message as well

### DIFF
--- a/src/helpers/validation.ts
+++ b/src/helpers/validation.ts
@@ -1,0 +1,36 @@
+import Ajv from 'ajv';
+import { isAddress } from '@ethersproject/address';
+
+export function validateForm(schema, form) {
+  const ajv = new Ajv({ allErrors: true });
+
+  ajv.addFormat('address', {
+    validate: isAddress
+  });
+
+  ajv.addFormat('uint256', {
+    validate: value => !!value.match(/^[0-9]+$/)
+  });
+
+  ajv.validate(schema, form);
+
+  const output = {};
+  if (!ajv.errors) {
+    return output;
+  }
+
+  for (const error of ajv.errors) {
+    const path = error.dataPath.split('.').slice(1);
+
+    let current = output;
+    for (let i = 0; i < path.length - 1; i++) {
+      const subpath = path[i];
+      if (!current[subpath]) current[subpath] = {};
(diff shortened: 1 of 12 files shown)


bonustrack:
Modal isn't showing up for me when I click on "Contract call" button, it fail with this error:
image

src/helpers/validation.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export function validateForm(schema, form) {
+  const ajv = new Ajv({ allErrors: true });
+
+  ajv.addFormat('address', {
+    validate: isAddress
+  });
+
+  ajv.addFormat('uint256', {
+    validate: value => !!value.match(/^[0-9]+$/)
```
bonustrack, line 12: There is a constant on ethers.js with the max value you can get, we could use this for validation https://docs.ethers.io/v5/api/utils/constants/#constants-MaxUint256, we could check if number is between 0 and max instead of using string filtering, this would enforce number to not start with zero(s) example: 00012345.
We could also add a format for `int256` which is same as `uint256` expect that it can go to negative there is also constants for the limits see: https://github.com/ethers-io/ethers.js/blob/master/packages/constants/src.ts/bignumbers.ts#L10-L11 you will need to update here too to add new format https://github.com/snapshot-labs/sx-ui/blob/db27536ad7ed659990d0c044ce2dc880f2e593a8/src/helpers/utils.ts#L115-L118

Sekhmet:
> Modal isn't showing up for me when I click on "Contract call" button, it fail with this error:

Does it still happen? I can't reproduce, I see there are some 530 HTTP errors in the console, maybe it failed to fetch something else and UI is broken :thinking:

bonustrack (approved):
Works well now!

Issue #184, opened 2022-06-27
Make it possible to drag and drop to change transaction order


bonustrack:
When creating a proposal with execution it should be possible to change the order of the transactions, we can add the icon "switch-vertical" (for lack of better icon) and make it drag and drop to sort transaction. See example:
<img width="636" alt="image" src="https://user-images.githubusercontent.com/16245250/175993021-387f3e07-42b2-4600-a859-874d2855f49a.png">
Ideally it should work on mobile too

Issue #183, opened 2022-06-27
Add nonce and operation on the execution transaction payload


bonustrack:
We should add on the execution transaction payload the params "nonce" which is the order in which the transaction should be executed, the first tx of an execution should have nonce=0 then second one get nonce=1 etc. And we should add a param "operation" which is always "0" for now but will be used in future to define delegateCall. Here is an example on how it would look like (skipping _form and _type metadata)
```json
{
  "transactions": [
    {
      "to": "0x9241627b130665b0a1e5636f1cc211895687bf25",
      "data": "0xf242432a00000000000000000000000010a19e7ee7d7f8a52822f6817de8ea18204f2e4f000000000000000000000000ef8305e140ac520225daf050e2f71d5fbcc543e7000000000000000000000000000000000000000000000000000000000000000b000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000",
      "value": "0",
      "nonce": 0,
      "operation": "0"
    },
    {
      "to": "0x9241627b130665b0a1e5636f1cc211895687bf25",
      "data": "0xf242432a00000000000000000000000010a19e7ee7d7f8a52822f6817de8ea18204f2e4f000000000000000000000000ef8305e140ac520225daf050e2f71d5fbcc543e7000000000000000000000000000000000000000000000000000000000000000b000000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000a000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000",
      "value": "0",
      "nonce": 1,
      "operation": "0"
    },
    {
      "to": "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
      "data": "0x",
      "value": "1000000000000000000",
      "nonce": 2,
      "operation": "0"
    }
  ],
  "title": "Hey yo",
  "body": ""
}
  ```

Sekhmet:
Do we need to persist it in drafts as well, or we can just add it once we submit it to the server? On the UI side `nonce` will be redundant as it will be basically index of the transaction in execution array so adding it there only complicates the thing. On the UI we have to use array order for it as that's how vuedraggable handles it.

bonustrack:
It's not necessary to persist it in the draft, it's just required when creating a proposal.

PR #212, opened 2022-07-17
feat: make executions sortable

Sekhmet:
Closes:  https://github.com/snapshot-labs/sx-ui/issues/184

## Test plan

- Create a new proposal (old drafts might not work because this PR adds unique IDs to executions).
- You can sort them.
- New sort order gets persisted on reload.

## Screenshots

https://user-images.githubusercontent.com/1968722/179423711-864aa58d-f29f-4400-9782-730c92809695.mov


bonustrack:
Here is my review:
- On mobile it's possible to drag but it's not possible to edit or delete a transaction for me, click don't trigger anything beside sort. We may need on small screen to always show the icons (drag, edit, delete) and make only drag icon activate the sorting. On normal screen we should make icons drag, edit, delete only appear on mouse over
- We shouldn't be able to drag if there is only 1 transaction
- What's the need of "id"? Could we use "nonce" instead? See related issue: https://github.com/snapshot-labs/sx-ui/issues/183

bonustrack:
Please check again my review I made some changes @Sekhmet

Sekhmet:
> What's the need of "id"? Could we use "nonce" instead? See related issue

Mostly for vuedraggable, but it looks like it was possible to workaround it by refactoring how we handle formatting title.

I think I covered all points from review.

bonustrack (approved):
Looks good

PR #222, opened 2022-07-23
feat: add nonce and operation to proposal file

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/183

- Added missing required `itemKey` prop, we don't have a good key to use for those, but we can use function to opt-out to default Vue keying.
- Nonce and operation are added when those are submitted to IPFS as those are derivatives or existing data in UI so it doesn't make sense to add it outside of it and would make it harder to work with (making sure we keep those in sync when we add/remove/reorder entries).

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -5,7 +5,7 @@ import { useWeb3 } from '@/composables/useWeb3';
 import { useTxStatus } from '@/composables/useTxStatus';
 import { useAccount } from '@/composables/useAccount';
 import { useModal } from '@/composables/useModal';
-import type { Transaction } from '@/types';
+import type { Transaction, TransactionData } from '@/types';
 
 export function useActions() {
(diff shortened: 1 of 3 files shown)


src/composables/useActions.ts
```diff
@@ -45,7 +45,21 @@ export function useActions() {
     execution: Transaction[]
   ) {
     if (!web3.value.account) return await forceLogin();
-    const pinned = await pin({ title, body, discussion, execution });
+
+    const executionData = execution.map(
+      (tx: Transaction, i: number): TransactionData => ({
+        ...tx,
```
Sekhmet, line 51: Do we want to remove `_type` and `form` from transaction in the stored file?

bonustrack: No we should keep it to better display them after

bonustrack (approved):
Sound good

Issue #178, opened 2022-06-25
Handle payable property in Transaction modal


Sekhmet:
We currently show value input for payable methods, but it's not connected to anything.

PR #223, opened 2022-07-23
feat: handle payable methods

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/178

## Test plan
- add contract call to `0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6`.
- `deposit` method has payable field, `transfer` doesn't.
- add `deposit` method call with amount.
- when editing the amount is there.
- log all `execution` it should contain `value` for calls with payable methods.


bonustrack (approved):
Looks good

Issue #188, opened 2022-07-03
Token picker, search by token address


bonustrack:
Currently it's not possible to search token per address on the token picker, we should add this.
<img width="568" alt="image" src="https://user-images.githubusercontent.com/16245250/177035152-06148877-d4b6-4e4b-b4b3-d121b8101877.png">

PR #228, opened 2022-07-23
feat: add contract address based search in token picker

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/188

## Test plan
- Searching for `0xba100000625a3754423978a60c9317c58a424e3D` should return BAL.

### DIFF
--- a/src/components/Block/TokenPicker.vue
+++ b/src/components/Block/TokenPicker.vue
@@ -23,6 +23,9 @@ const filteredAssets = computed(() =>
           .includes(props.searchValue.toLocaleLowerCase()) ||
         asset.contract_name
           .toLocaleLowerCase()
+          .includes(props.searchValue.toLocaleLowerCase()) ||
+        asset.contract_address
+          .toLocaleLowerCase()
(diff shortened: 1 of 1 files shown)


src/components/Block/TokenPicker.vue
```diff
@@ -22,6 +22,9 @@ const filteredAssets = computed(() =>
           .toLocaleLowerCase()
           .includes(props.searchValue.toLocaleLowerCase()) ||
         asset.contract_name
+          .toLocaleLowerCase()
+          .includes(props.searchValue.toLocaleLowerCase()) ||
+        asset.contract_address
```
Sekhmet, line 27: It's unlikely that anyone would type the address characater by character so we could limit it to an exact match (`asset.contract_address === props.searchValue`).

bonustrack: Yes agree, it's better to do exact match here to avoid collision with others token symbol

bonustrack (approved):
TACK

Issue #185, opened 2022-06-29
Add contact list helper on execution modals address inputs


bonustrack:
It could be visible on the recipient input like this:
<img width="547" alt="image" src="https://user-images.githubusercontent.com/16245250/176405957-66bf3c82-2e16-4940-86bc-35821e5a8b28.png">
This is using the icon "identification". If you click on the icon it would send you to a modal similar than token picker but instead you will pick in a defined list, the list could container the ERC20 example that we had on the contract call modal, we could also add current logged address in the list. We shouldn't add the icon in the S/ input components, we should keep this components general purpose.

Need design for the contact picker

PR #242, opened 2022-07-24
feat: add ContactPicker

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/185

Long term I think it would make sense to create a more robust modal system that can render them at will, outside of main tree so we could just do `const result = await showContactPicker()`. That would improve reusability of modals a lot (right now they need to repeat some of the UI). I have no idea how to do it now, seems common with React/Redux, but for Vue it seems it's never done like that.

## Test plan
- Open different modal types, you can pick address in there.
- Open Contract modal.
- Pick WETH contract and `transferFrom` method.
- You can pick different addresses from picker.

## Screenshots

image

### DIFF
--- a/src/components/Block/ContactPicker.vue
+++ b/src/components/Block/ContactPicker.vue
@@ -1,8 +1,6 @@
 <script setup lang="ts">
 import { computed } from 'vue';
-import { formatUnits } from '@ethersproject/units';
-import { ETH_CONTRACT } from '@/helpers/constants';
-import { _n, shorten } from '@/helpers/utils';
+import { shorten } from '@/helpers/utils';
 
 const CONTACTS = [
   {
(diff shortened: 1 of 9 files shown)


bonustrack:
@Sekhmet 

The UI looks good and yes we want it on the contract call modal too. 

For the generated form it make sense to have the contact picker for address fields but we want to make sure the S/ components stay reusable and unopinionated, I don't think we should add the icon or modal interaction in the S/ components. Maybe a solution would be to use `<slot/>` or create custom component `Ui/IAddress` that wrap `S/IString`.

For the modal system or anything reducing code duplicate I'm all for it, feel free to make an issue if you have idea on how to improve it.

Sekhmet:
@bonustrack this should be ready for review.

bonustrack:
Can we add in the contact list the address of the current user if he is logged? We can label it "You".

bonustrack (approved):
tACK

PR #269, opened 2022-08-13
fix: make sx-ui compatible with latest sx.js

Sekhmet:
Closes: https://github.com/snapshot-labs/sx.js/issues/43

New ethereumjs libs need `stream` and `Buffer` to be present, which are node.js deps. We need to polyfill those in the browser.

## Test plan
- `yarn && yarn dev`
- UI loads without errors


bonustrack:
I just see this increase the app build by like ~20%, I don't think we directly use `ethereumjs` in the UI, isn't there a better way?
Before https://goerli.snapshotx.xyz/stats.html
After https://bafybeidbql47zmaqpcufqxqryl4dvdjwtrxy3uwgtioxx2ora2wtetdeoq.on.fleek.co/stats.html

bonustrack:
@Sekhmet Actually that library is not used atm but will most likely be used later so you no need to bother.
Another thing, deployment on with Fleek show blank screen there is the error "uncaught ReferenceError: Buffer is not defined" see: https://bafybeidbql47zmaqpcufqxqryl4dvdjwtrxy3uwgtioxx2ora2wtetdeoq.on.fleek.co/

Sekhmet:
@bonustrack thank you for catching that, fixed it.

bonustrack (approved):
LG

PR #277, opened 2022-08-20
feat: self host Space Mono font

### LINKED ISSUE #270: Add font source instead of loading from external
Google font get catch by ad blocker we should instead add the font source directly in the code and avoid external calls.
image

Sekhmet:
Closes #270

## Test plan
- Font still loads (from our own server).


bonustrack (approved):
LG

PR #311, opened 2022-09-23
feat: use latest @snapshot-labs/sx

Sekhmet:
Closes: #288

I hardcoded executor + authenticator for now. We will need to update `sx-api` to return those in better format (right now it's string-encoded array. And it would be nice to use some state (Vuex) to persist spaces/proposals so we can easily access them when we need to vote/propose.

## Test plan

Use this mana: https://github.com/snapshot-labs/mana/pull/79

Use ethSig spaces for now as authenticator is hardcoded.

- Create proposal
- Vote on proposal

### DIFF
--- a/src/helpers/clients.ts
+++ b/src/helpers/clients.ts
@@ -1,8 +1,13 @@
 import { clients as Clients } from '@snapshot-labs/sx';
 
-const url: string = import.meta.env.VITE_MANA_URL || 'http://localhost:3000';
+const manaUrl: string =
+  import.meta.env.VITE_MANA_URL || 'http://localhost:3000';
+const ethUrl: string = import.meta.env.VITE_ETH_NODE_URL;
 
 export default {
-  EthereumSig: new Clients.EthereumSig(url),
-  StarkNetSig: new Clients.StarkNetSig(url)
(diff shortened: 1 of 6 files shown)


bonustrack:
For state management we may want use Pinia

bonustrack:
Somehow, when I try to cast a vote my wallet (MetaMask) is not triggered, I can see requests are sent to L2 node but there isn't any error or message in the console

image

image

src/helpers/clients.ts
```diff
@@ -1,8 +1,13 @@
 import { clients as Clients } from '@snapshot-labs/sx';
 
-const url: string = import.meta.env.VITE_MANA_URL || 'http://localhost:3000';
+const manaUrl: string =
+  import.meta.env.VITE_MANA_URL || 'http://localhost:3000';
+const ethUrl: string = import.meta.env.VITE_ETH_RPC_URL;
```
bonustrack, line 5: Could you add this in the .env file:
```VITE_ETH_RPC_URL=https://rpc.ankr.com/eth_goerli```
Just to make it simpler for contributors to get setup

bonustrack (approved):
Had an invalid nonce previously but otherwise it works well

PR #314, opened 2022-09-24
chore: enforce prettier to pass

Sekhmet:
- Make prettier ruler errors.
- Update eslint path so it runs on files directly in `src` too.
- Reformat files that are not properly formatted.


bonustrack (commented):
LGTM

bonustrack (approved):

PR #320, opened 2022-09-25
feat: add Vote component

Sekhmet:
This PR adds Vote component that unifies logic for handling voting edge cases:
- User already voted for proposal
- Proposal is not supported (unsupported authenticators/strategies).

With this component you just provide different UI for those cases so you don't have to worry that in some place you can vote when you shouldn't.

This PR also makes Vote ratio to show up again.

## Test plan

- In Space view, proposals that were already voted on have vote ratio displayed.
- In Space view, proposals that are unsupported (go to Tripod DAO) don't have buttons to vote.
- In Space view, other proposals have vote buttons visible.
- In Proposal view, proposals that were already voted have "You have already voted for this proposal" message.
- In Proposal view, proposals that are unsupported (go to Tripod DAO) have "Voting for this proposal is not supported" message.
- In Proposal view, other proposals have vote buttons visible.

## Screenshots

image
image

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,8 @@
 module.exports = {
   root: true,
   env: {
-    node: true
+    node: true,
+    'vue/setup-compiler-macros': true
   },
   extends: [
     'plugin:vue/vue3-recommended',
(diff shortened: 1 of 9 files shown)


Sekhmet:
Drafting until sx-api is updated to handle subqueries instead.

bonustrack (commented):
Looks good, just some suggestions:
- Could we show the results on the proposal page if the user has voted?
- We could also show results and hide buttons everywhere if the proposal is closed (you'll need to compare current time to proposal `max_end` param).

bonustrack (approved):
Perfect

PR #321, opened 2022-09-25
feat: use authenticator/strategies from API

Sekhmet:
Depends on https://github.com/snapshot-labs/sx-ui/pull/320

## Test plan

- Create proposal.
- Vote on proposal.


bonustrack (approved):
utACK

PR #323, opened 2022-09-27
feat: support sx-api with subqueries

Sekhmet:
Test with: https://github.com/snapshot-labs/sx-api/pull/152

## Test plan
- Whole app should work as before.

### DIFF
--- a/src/components/Modal/Votes.vue
+++ b/src/components/Modal/Votes.vue
@@ -1,18 +1,21 @@
-<script setup>
-import { ref, toRefs, watch } from 'vue';
+<script setup lang="ts">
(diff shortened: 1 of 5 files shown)


bonustrack (commented):
Query to get votes is failing for me
image

bonustrack:
Changing to `props.proposal.space.id` here does the fix but there is still some issue to render the modal content. 
https://github.com/snapshot-labs/sx-ui/blob/f9e5550097c81a40ebb55b98983c8e9ee08b0f81/src/components/Modal/Votes.vue#L25

bonustrack (approved):
Looks good!

Issue #326, opened 2022-10-02
Results bars based on quorum


bonustrack:
Currently we use the value `scores_total` to define the full width of the results bars. Instead of having something like this:
<img width="584" alt="image" src="https://user-images.githubusercontent.com/16245250/193479746-f27e0380-7b06-47fb-88ec-b8cc3c9df0d4.png">

We would have something like this:
<img width="583" alt="image" src="https://user-images.githubusercontent.com/16245250/193479700-d824d7fd-e861-4ca7-8063-ebed902df62c.png">
The darker grey space is what's missing to get the quorum on the proposal.

PR #328, opened 2022-10-09
feat: adjust results using quorum

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/326

## Test plan

Test against as Vanilla strategy gives super small scores as sx-api formats them as amounts.

## Screenshots

image

### DIFF
--- a/src/components/Results.vue
+++ b/src/components/Results.vue
@@ -1,7 +1,8 @@
 <script setup lang="ts">
+import { computed } from 'vue';
 import { Proposal as ProposalType } from '@/types';
 
-withDefaults(
+const props = withDefaults(
   defineProps<{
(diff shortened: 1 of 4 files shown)


bonustrack:
Looks good! Does this handle the case when there is more with for + "against + abstain than the quorum? In this case we should assume full width is for + against + abstain.

Sekhmet:
@bonustrack `Math.min` there was supposed to handle that:
https://github.com/snapshot-labs/sx-ui/blob/f99cf51a1b49cb28e992e5da75a7d5f1297399a7/src/components/Results.vue#L22

but I just noticed it should have been `Math.max` :D

bonustrack (approved):
tACK

Issue #325, opened 2022-10-02
Add section with closed proposals on space overview


bonustrack:
We should have 2 sections in the space overview page, one with active proposal and another with closed proposal like this:
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/16245250/193474452-10733cae-848b-4955-8a6b-a6ab22264b21.png">
We can limit the number of proposal visible to 3 per section. The links "See more" can just redirect to the space proposals page.

PR #330, opened 2022-10-09
feat: add Closed proposals section

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/325

## Screenshots

image

### DIFF
--- a/src/components/ProposalsList.vue
+++ b/src/components/ProposalsList.vue
@@ -1,0 +1,38 @@
+<script setup lang="ts">
+import type { Proposal as ProposalType } from '@/types';
+
+defineProps<{
+  title: string;
+  loading?: boolean;
+  proposals: ProposalType[];
+  route?: {
+    name: string;
+    linkTitle: string;
+  };
+}>();
+</script>
+
+<template>
+  <div>
+    <Label :label="title" />
+    <UiLoading v-if="loading" class="block px-4 py-3" />
+    <div v-else>
+      <Proposal
+        v-for="(proposal, i) in proposals"
+        :key="i"
+        :proposal="proposal"
+      />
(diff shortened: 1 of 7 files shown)


src/components/ProposalsList.vue
```diff
@@ -0,0 +1,46 @@
...
+  route?: {
+    name: string;
+    linkTitle: string;
+  };
+}>();
+
+const limit = computed(() => {
+  if (props.limit === false) return Infinity;
+
+  return props.limit || 3;
```
bonustrack, line 19: It seem that the limit of 3 doesnt work for me, I can see 4 closed proposals on this space:
https://bafybeibh4naq5yi67sa43jmlpoojvh646f5oyrsc6fgqd7h6lcpy7i4gd4.on.fleek.co/#/0x04b7cff71219e275676e0ca23579f41b99dd1d1bd01adc7d7f1bc917d448e57d

bonustrack (commented):
This works but we may not want to rely on the UI to filter the proposals for too long, it will be more tricky when there is more proposals than what we query in a single request. We could query the active and closed proposals from the API, it will probably break Pinia, we wont be able to reuse the same data from the overview page in the proposals page, not sure there is solution for this.

src/views/Space/Overview.vue
```diff
@@ -1,19 +1,37 @@
 <script setup lang="ts">
 import { onMounted, computed } from 'vue';
 import { useProposalsStore } from '@/stores/proposals';
-import { Space } from '@/types';
+import { Space, Proposal as ProposalType } from '@/types';
+
+const PROPOSALS_LIMIT = 4;
```
Sekhmet, line 6: We need to fetch one more than we want to display to be able to detect if we need See more link.

bonustrack (approved):
Works well

Issue #315, opened 2022-09-24
Move to OpenSea API to query NFTs


bonustrack:
Currently we are using Covalent but it's often slow to resolve. Moving to OpenSea would give us more rich context (like price) and probably a more complete listing.

Sekhmet:
@bonustrack API to [list NFTs](https://docs.opensea.io/reference/getting-assets) requires API key, can you [request API](https://docs.opensea.io/reference/request-an-api-key) key from OpenSea? Looks like they have manual approval process and doing this though Snapshot's email might get us the key faster.

bonustrack:
@Sekhmet I've made a request 2 weeks ago with Snapshot email but got no reply, on their Discord there is many others in same situation, we should get the key at some point I'll do another request if no progress. This should not be a blocker for the integration, their API doesn't require API key if you use their testnet which is on Goerli, for us it will be ideal since our testnet is also on Goerli.

PR #331, opened 2022-10-09
feat: use OpenSea for fetching NFTs

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/315

## Screenshots

image

### DIFF
--- a/src/components/NftPreview.vue
+++ b/src/components/NftPreview.vue
@@ -1,19 +1,27 @@
 <script setup>
+import { computed } from 'vue';
+
 const props = defineProps({
   item: Object,
   size: Number
 });
 
-function handleError(e) {
-  const size = props.size ? props.size * 2 : 256;
+const fallbackUrl = computed(
+  () =>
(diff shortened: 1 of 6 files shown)


bonustrack (commented):
Nice! We may want to also truncate the NFT name in the selector to get consistent item height
image

src/components/NftPreview.vue
```diff
@@ -1,19 +1,27 @@
...
 const props = defineProps({
   item: Object,
   size: Number
 });
 
-function handleError(e) {
-  const size = props.size ? props.size * 2 : 256;
+const fallbackUrl = computed(
+  () =>
+    `https://cdn.stamp.fyi/token/${props.item.contractAddress}?s=${
```
bonustrack, line 11: We can use the Stamp NFT endpoint in future to avoid fallback

Sekhmet: That PR wasn't merged yet:
https://github.com/snapshot-labs/stamp/pull/28

bonustrack: Yep I'll make review there

PR #334, opened 2022-10-11
refactor: Change treasury address to Goerli owned Safe

bonustrack:
Changing the hardcoded treasury address to the space L1 controlled Safe, this is to make it easier to test execution later.

### DIFF
--- a/src/views/Space/Treasury.vue
+++ b/src/views/Space/Treasury.vue
@@ -31,7 +31,7 @@ const sortedAssets = computed(() =>
 );
 
 onMounted(() => {
-  loadBalances(space.wallet);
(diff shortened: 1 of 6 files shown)


src/views/Space/Treasury.vue
```diff
@@ -88,16 +88,9 @@ onMounted(() => {
             <div class="flex flex-col ml-3 leading-[22px]">
               <h4
                 class="text-skin-link"
-                v-text="
```
bonustrack, line 91: Also made a fix here, we should show asset symbol and name

Sekhmet (approved):
utACK

Issue #333, opened 2022-10-11
Display execution in the proposal page


bonustrack:
We should show if a proposal have execution, and it should look consistent with the execution section in the editor page, like this:

<img width="673" alt="image" src="https://user-images.githubusercontent.com/16245250/195058614-be3f08a8-babb-4c84-9c8f-4adbdb4576f7.png">

Ideally we should avoid code duplicate

PR #335, opened 2022-10-11
feat: add execution to proposal page

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/333

## Test plan

- Check this proposal: `#/0x04b7cff71219e275676e0ca23579f41b99dd1d1bd01adc7d7f1bc917d448e57d/proposal/8`

## Screenshots

image

### DIFF
--- a/src/views/Proposal.vue
+++ b/src/views/Proposal.vue
@@ -1,10 +1,11 @@
 <script setup lang="ts">
-import { onMounted, ref, computed } from 'vue';
+import { onMounted, ref, computed, watch, Ref } from 'vue';
 import { useRoute } from 'vue-router';
 import { sanitizeUrl } from '@braintree/sanitize-url';
 import { useProposalsStore } from '@/stores/proposals';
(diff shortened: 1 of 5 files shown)


src/views/Proposal.vue
```diff
@@ -25,7 +28,32 @@ const discussion = computed(() => {
...
 });
 
+const metadataUrl = computed(() => {
+  if (!proposal.value) return null;
+
+  return getUrl(proposal.value.metadata_uri);
+});
+
+async function fetchMetadata() {
+  const res = await fetch(metadataUrl.value);
```
bonustrack, line 38: Is this just to get the execution data? It would be better to do index it on sx-api to avoid extra request to IPFS in the frontend. These request can be slow

Sekhmet: Thanks. I didn't even notice it was already in sx-api.

bonustrack (approved):
tACK

Issue #336, opened 2022-10-14
Bug with discussion section on small screen


bonustrack:
The discussion section overflow on small screen, see here:
![Screenshot_20221014-181632_Chrome](https://user-images.githubusercontent.com/16245250/195834391-4fd63ce2-87a1-4aac-a8fe-4967adfb74a1.jpg)
We should make it so the text truncate to remove overflow, we should also get border back on left and right.

PR #339, opened 2022-10-15
fix: prevent too-wide UI

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/336

## Screen shots

image


bonustrack:
For the discussion block could you add the left and right border and radius on small screen? Make it same as execution block
image

Also there seem to be something wrong on the right side, the space in left and right should be same width but currently its not, see here:
image

Compared to GitHub
image

bonustrack (approved):
Perfect!

PR #344, opened 2022-11-25
chore: format codebase with shared config

Sekhmet:
## Summary

Reformat codebase using `@snapshot-labs/prettier-config` and `@snapshot-labs/eslint-config-vue`.
Change in files is result of different printWidth in new config (100 vs old 80).

## Test plan

CI should pass.


bonustrack (approved):
utACK

Issue #343, opened 2022-11-24
Fix WalletConnect 


bonustrack:
For some reason WalletConnect is not working anymore here is the error:
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/16245250/203772041-225ee75b-221e-4037-8002-d97bb9a9d19f.png">

PR #346, opened 2022-12-06
fix: convert mixed ESM

Sekhmet:
Closes https://github.com/snapshot-labs/sx-ui/issues/343

Starting from https://github.com/snapshot-labs/sx-ui/commit/f71b654b89525eeb74cadc79523f786d254ed57f we are using `@walletconnect/web3-provider` in newer version that added ESM module, but it has mixed syntax (both import and require) and it needs to be converted before it can be used.

## Test plan

- Click Connect -> WalletConnect.
- WalletConnect window should open.
- Test both in `yarn dev` and `yarn build` if possible.

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -50,6 +50,9 @@ export default defineConfig({
     }
   },
   build: {
+    commonjsOptions: {
+      transformMixedEsModules: true
+    },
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #348, opened 2022-12-09
feat: compute ethRelayer execution

Sekhmet:
## Summary

This PR makes execution through ethRelayer functional. As of now I only tested sending ETH out.

## Test plan

- Create proposal with ETH transfer as execution.

For full test plan with execution check https://github.com/snapshot-labs/sx.js/pull/115 as finalizing and executing proposal isn't part of the UI.

## Further comments

There seems to be something wrong with Vite setup after previous PRs, if you have issues opening the app you can copy `dist` instead of linking.

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -5,7 +5,7 @@ import { useWeb3 } from '@/composables/useWeb3';
 import { useTxStatus } from '@/composables/useTxStatus';
 import { useAccount } from '@/composables/useAccount';
 import { useModal } from '@/composables/useModal';
-import type { Transaction, TransactionData, Proposal, Space } from '@/types';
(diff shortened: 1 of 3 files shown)


bonustrack:
@Sekhmet Got this error when trying to create a proposal from Fleek deployment
image
And tested locally with yarn link to https://github.com/snapshot-labs/sx.js/pull/115 but got that error
image
Am I missing something?

Sekhmet:
@bonustrack should work now, it was assuming https://github.com/snapshot-labs/sx.js/pull/114 was merged, but master has other changes that are not compatible. Added different imports paths to make it work with current `sx` package. Will update usage here once new version is released.

bonustrack (approved):
tACK

PR #349, opened 2022-12-14
feat: abstract away API calls to fetch data

Sekhmet:
## Summary

This PR centralizes API calls into single place instead of having it scattered over the codebase.


bonustrack (approved):
tACK

PR #351, opened 2022-12-15
refactor: move actions to network

Sekhmet:
## Summary

This PR moves actions to common file, next to API calls instead of having them in UI actions.

## Test plan

- create proposal
- vote


bonustrack (approved):
tACK

PR #352, opened 2022-12-15
fix: prevent getting stuck at editor

Sekhmet:
## Summary

Currently pressing back button in browser at editor screen won't make you go back. Also deleting the draft doesn't make you go back to fresh editor and instead leaves you at current URL

## Test plan

- Go to editor through UI.
- Press back.
- You get back to space.


bonustrack (changes requested):
If I delete a draft the scroll of the page get broken, can you reproduce?

bonustrack (approved):
tACK

Issue #353, opened 2022-12-19
Going back from Proposals to Overview leaves "Closed proposals" empty


Sekhmet:
![image](https://user-images.githubusercontent.com/1968722/208418028-3a515647-fc9d-44b4-adae-f2a4b6977e2c.png)

PR #356, opened 2022-12-19
fix: separate summary from main proposals

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/353

Summary has its own logic how it's fetched so using main proposals as substitute for summary won't work as we won't have enough old proposals in the end. Trying to handle it somehow will be mess so I'm just separating them into two different properties.

## Test plan
- Go to http://127.0.0.1:8080/#/0x07e6e9047eb910f84f7e3b86cea7b1d7779c109c970a39b54379c1f4fa395b28/proposals
- Click Overview - it fetches as expected.
- Go back to Proposals - no duplicates (but doesn't fetch again).
- Click on Proposal from Summary/Proposals page - it doesn't refetch it.

### DIFF
--- a/src/stores/proposals.ts
+++ b/src/stores/proposals.ts
@@ -6,21 +6,26 @@ import type { Proposal } from '@/types';
 type SpaceRecord = {
   loading: boolean;
   loaded: boolean;
+  proposals: Proposal[];
   summaryLoading: boolean;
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

Issue #189, opened 2022-07-03
Token picker, load token by address


bonustrack:
It should be possible to import a token that is not on the list in token picker modal. To do so we will need to detect when someone type an address on the search and if there isn't any result we can try to load token information using a multicall. Here is an example of multicall to get a token name, symbol and decimals:
https://github.com/snapshot-labs/snapshot.js/blob/master/test/multicaller.ts
We should also query `balanceOf` to get treasury wallet balance

PR #358, opened 2022-12-19
feat: add AddToken picker

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/189
Depends on https://github.com/snapshot-labs/sx-ui/pull/361

## Test plan

- Co to editor, click Send token.
- Open token picker.
- Paste ERC20 into search (`0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6`).
- Pick it as token.
- Add execution with it.


bonustrack:
@Sekhmet Works well, but I think the UX should be slightly different, instead of having a button "add a token" we should make it so you can use the search input to load new token. Pretty much what Uniswap is doing, you can try Uniswap on Goerli and type "0xf7a702d8f197e6d510eaf740998c2029744078b5" in the search to see it (we don't need the modal with warning).

bonustrack:
When looking for a token that doesn't exist for example `0xb4fbf271143f4fbf7b91a5ded31805e42b2208d2` we should show a message "No results", currently it doesn't show anything 
image
Also we should remove this double border here 
image

bonustrack (approved):
tACK

Issue #172, opened 2022-06-22
Move to Alchemy API?


bonustrack:
For some reason sometime requests to Covalent API are slow to resolve, it might be more stable / fast if we move to Alchemy API https://docs.alchemy.com/alchemy

bonustrack:
For NFTs we could use OpenSea API directly too

PR #361, opened 2022-12-19
feat: migrate balances to Alchemy

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/172

Moves fetching token balances to Covalent. Pricing data is missing (which was also missing on Goerli with Covalent). We will need to use CMC/CG to add pricing info.

## Test plan

- Check treasury page.
- Check Send Token token picker.


bonustrack (approved):
tACK

PR #362, opened 2022-12-20
feat: add infinite scroll to proposals

Sekhmet:
## Summary

- Add `InfiniteScroller` component.
- Split `fetchAll` into `fetch` and `fetchMore` (a bit more code repeating, but putting all logic for both first call and more fetches in single function was making it a bit hard to understand).
- Convert proposals page to be infinite scrollable.

## Test plan

- Go to http://127.0.0.1:8080/#/0x07e6e9047eb910f84f7e3b86cea7b1d7779c109c970a39b54379c1f4fa395b28/proposals
- You can fetch whole list.
- Nothing else breaks.

### DIFF
--- a/src/stores/proposals.ts
+++ b/src/stores/proposals.ts
@@ -5,12 +5,16 @@ import type { Proposal } from '@/types';
 
 type SpaceRecord = {
   loading: boolean;
+  loadingMore: boolean;
   loaded: boolean;
   proposals: Proposal[];
+  hasMoreProposals: boolean;
   summaryLoading: boolean;
   summaryLoaded: boolean;
   summaryProposals: Proposal[];
 };
+
+const PROPOSALS_LIMIT = 20;
 
 export const useProposalsStore = defineStore('proposals', {
   state: () => ({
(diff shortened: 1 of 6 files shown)


src/stores/proposals.ts
```diff
@@ -5,13 +5,17 @@ import type { Proposal } from '@/types';
...
+  loadingMore: boolean;
   loaded: boolean;
   proposals: Proposal[];
+  hasMoreProposals: boolean;
   summaryLoading: boolean;
   summaryLoaded: boolean;
   summaryProposals: Proposal[];
 };
 
+const PROPOSALS_LIMIT = 20;
```
bonustrack, line 17: Loading 20 proposals is a bit too much it create large scrollbar can we change to 10?
```suggestion
const PROPOSALS_LIMIT = 10;
```

Sekhmet: I guess doing just 10 is a bit small for infinite scroll (especially as elements are tiny and with small scroll you would already end up with fetching more).
If we really want to make it lower, maybe we can just lower it for initial load (10), but when fetching more we will fetch 20 again?

bonustrack: Even if elements are tiny, with 10 of them it already fit more than a screen page on large screen so I'm not sure it worth loading 20. I don't think we should bother adding more complexity with fetching 10 then 20. But not a strong opinion, I just looked at Twitter and they loading quite a lot of tweet on first load, we can go with 20.

bonustrack (approved):
tACK

Issue #357, opened 2022-12-19
Implement pagination for proposals and spaces


Sekhmet:
Currently we only fetch N latest proposals and spaces on UI. We should add pagination for it.

PR #363, opened 2022-12-21
feat: add infinite scroll to spaces

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/357

This PR adds infinite scroll to spaces on front page. It's disabled by default and triggers only when user presses "Load more" button, similar as on Snapshot.org

## Test plan
- No Button to "Load more" is visible.
- Change `SPACES_LIMIT` to 2.
- Go to homepage
- Button shows up, click it.
- New elements are loaded.

### DIFF
--- a/src/networks/starknet/queries.ts
+++ b/src/networks/starknet/queries.ts
@@ -170,8 +170,8 @@ export const SPACE_QUERY = gql`
 `;
 
 export const SPACES_QUERY = gql`
-  query {
-    spaces {
+  query ($first: Int!, $skip: Int!) {
+    spaces(first: $first, skip: $skip, orderBy: created, orderDirection: desc) {
       id
       name
       about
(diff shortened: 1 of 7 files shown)


src/networks/starknet/queries.ts
```diff
@@ -170,8 +170,8 @@ export const SPACE_QUERY = gql`
 `;
 
 export const SPACES_QUERY = gql`
-  query {
-    spaces {
+  query ($first: Int!, $skip: Int!) {
+    spaces(first: $first, skip: $skip, orderBy: created, orderDirection: desc) {
```
bonustrack, line 174: Looks good! Can you just order by vote_count here?
```suggestion
    spaces(first: $first, skip: $skip, orderBy: vote_count, orderDirection: desc) {
```

bonustrack (approved):
tACK

Issue #365, opened 2023-01-02
feat: sidebar menu on mobile


bonustrack:
We should add a button to toggle the sidebar on mobile. For the icon to toggle we can use `menu-alt-2` from Heroicons v1. 
Here is an example on how it would look like this:

<img width="841" alt="image" src="https://user-images.githubusercontent.com/16245250/210281014-7b51c34a-3492-462d-aeb8-4209db955c76.png">

The icon should not be visible on large screen. When user click on menu icon, the sidebar should open without animation and look like this:

<img width="879" alt="image" src="https://user-images.githubusercontent.com/16245250/210281923-695f55bc-cd57-4fa0-83ca-5295a64cb20b.png">

When the sidebar menu is open, the main content in right side (A) should have a background similar than the backdrop when we open a modal. If we click on the main content (A) it should collapse the sidebar menu.

Ideally we should be able to slide on mobile with finger from left to right to open or collapse the menu.

PR #368, opened 2023-01-06
feat: add responsive sidebar

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/365

With our quite complex structure of sidebars (we have 3 fixed elements, that all need to be shifted in different way) it ended up with a bit ugly set up as each of those has to contain logic for how it's going to be shifted (can't be shifted at parent level as transform modifies origin and `fixed` positioning no longer works.

Overall it should work as it used to on bigger screen (with exception of sidebar and nav now collapsing at once to have consistent experience). When on smaller screens those are assumptions that should be the case:
- You can use backdrop to dismiss.
- When picking element from sidebars sidebar closes.
- With sidebar open page is not scrollable.
- Content is translated right, instead of shrunk down (so it should look as before, but just moved out of the view).
- There is no swipe left-to-right to pull up the menu (so it doesn't break native swipe to go back gestures).

## Video

https://user-images.githubusercontent.com/1968722/211068094-6256e504-f6e9-4c2b-8c66-49b91885d84e.mp4

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,11 +1,15 @@
 <script setup>
 import { computed, onMounted, provide, watch } from 'vue';
+import { useRoute } from 'vue-router';
+import { useUiStore } from '@/stores/ui';
 import { useModal } from '@/composables/useModal';
 import { useUserSkin } from '@/composables/useUserSkin';
 import { useApp } from '@/composables/useApp';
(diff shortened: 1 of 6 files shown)


bonustrack (approved):
tACK

PR #369, opened 2023-01-09
fix: update Create page navbar position

Sekhmet:
## Summary

Create page navbar currently displays incorrectly at certain breakpoints. This PR adjust this behavior.

## Test plan

- Go to Create page
- Resize it from small to big, no weird stuff happens.

## Screenshots

Before:
image

After:
image

## Further notes

This screen doesn't show menu icon at the moment. I don't think it's a big deal, because you have back button still, but if we want to make menu accessible on this screen as well we might need to consider some design changes (right now showing both hamburger icon and back icon will look weird I think).

### DIFF
--- a/src/views/Create.vue
+++ b/src/views/Create.vue
@@ -34,7 +34,7 @@ const space = computed(() => spacesStore.spacesMap.get(id));
 </script>
 <template>
   <div>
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

Issue #367, opened 2023-01-02
feat: improve pending transaction indicator


bonustrack:
When someone do an action on the UI we should check with a node if the tx hash is confirmed at regular interval to confirm the tx on the UI. We can use StarkNet.js `waitForTransaction` function to do that.

We should also store the pending transaction hashes on local storage to be able to show pending transactions even if the user reload the page.

The pending transactions indicator can be moved to the topnav like this:

<img width="417" alt="image" src="https://user-images.githubusercontent.com/16245250/210283734-29de9356-3d35-4f68-a769-57aec9221c8c.png">

PR #371, opened 2023-01-10
fix: properly display topnav in space screen

Sekhmet:
## Summary

There were some issues with topnav 1) being translated on large screen causing button on right side to be off-screen 2) misaligned if sidebar was opened and the window was later resized to large size.

## Screenshots

After:

https://user-images.githubusercontent.com/1968722/211562700-582e59cb-88e1-4588-9cfc-8d474e776496.mp4

Before:

https://user-images.githubusercontent.com/1968722/211562901-239c3909-0ec3-4756-9896-aa7d51505ba0.mp4

PR #372, opened 2023-01-11
feat: add PendingTransactions indicator

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/367
Depends on: https://github.com/snapshot-labs/sx-ui/pull/371

This PR changes how pending transaction indicator works:
- It's now part of top nav, it handles both transactions that wait to be broadcast (same as old one) as well as track transactions until confirmation.
- Once txId is received you can browse pending transactions and open them in explorer.
- txIds are persisted.

## Test plan

- Propose/vote.
- Indicator gets increased once you confirm transaction in Metamask.
- You can click on indicator (shows X transactions broadcasting).
- Once response comes back it shows transaction ID and you can open it in explorer.
- Once it gets confirmed it gets removed.


bonustrack:
The space in red here is too large now:
image

Should be like before:
image

Can you also remove the white border here:
image

bonustrack:
Still seeing the white border for some reason

bonustrack:
- Is it supposed to persist with localStorage? I can't see the pending tx anymore when I refresh the page.
- Can you make the text white for pending count on light mode here:
image

Sekhmet:
> Is it supposed to persist with localStorage? I can't see the pending tx anymore when I refresh the page.

If you wait til transaction receipt is returned then it should. Creating proposal takes 20-30 seconds to receive tx_id, so if you refresh during that window we have no txId to track.

bonustrack (approved):
tACK

PR #394, opened 2023-01-30
fix: only show drafts from same space

Sekhmet:
## Summary

Currently we show drafts from all spaces, but if you try to open draft from different space it won't work (data won't be loaded).

We have two ways of fixing this:
- Show all drafts in any space, just navigate to different space when selected - I started with this solution initially, but thought it will be confusing to suddenly end up in different space. I think it makes sense that drafts are space-scoped.
- Only show drafts from current space (this PR does that)

## Test plan

- Remove all drafts
- Go to A space, create a draft
- Go to B space, no drafts in drafts modal, create some draft in space B.
- Go back to A space, you can only see draft created for space A, space B's drafts are not visible.

### DIFF
--- a/src/components/Modal/Drafts.vue
+++ b/src/components/Modal/Drafts.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { toRefs } from 'vue';
+import { toRefs, computed } from 'vue';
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

PR #396, opened 2023-01-30
fix: properly connect with ArgentX

Sekhmet:
## Summary

ArgentX wasn't working for me. On first connect nothing would happen, on second attempt it might work. Looks like it's because we don't actually call `connect`, but deprecated `getStarknet()`.

## Test plan

- Connect with ArgentX.
- It works.


bonustrack (approved):
tACK

PR #398, opened 2023-01-31
feat(starknet): make edit space functional

Sekhmet:
## Summary

Depends on https://github.com/snapshot-labs/sx.js/pull/128
Closes https://github.com/snapshot-labs/sx-ui/issues/378

This makes edit space functional for Starknet spaces

## Test plan

Sign in with ArgentX, edit space you control (you can use follow up PR to create new one if you need).

PR #400, opened 2023-02-01
feat(starknet): add simplified deploy space form

Sekhmet:
## Summary

Depends on https://github.com/snapshot-labs/sx-ui/pull/398
Depends on https://github.com/snapshot-labs/sx.js/pull/131

This adds very simplified UI to create Starknet spaces.

## Test plan

Sign in with ArgentX, click plus button in Topnav, fill the form and submit it.

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -33,6 +33,37 @@ export function useActions() {
     modalAccountOpen.value = true;
   }
 
+  async function createSpace(networkId: NetworkID, metadata: SpaceMetadata) {
+    if (!web3.value.account) return await forceLogin();
+
+    const network = getNetwork(networkId);
+    if (!network.managerConnectors.includes(web3.value.type as Connector)) {
+      throw new Error(`${web3.value.type} is not supported for this actions`);
+    }
+
+    const pinned = await network.helpers.pin(metadata);
+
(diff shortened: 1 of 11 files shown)


src/composables/useActions.ts
```diff
@@ -33,6 +41,44 @@ export function useActions() {
...
+
+    const pinned = await network.helpers.pin(metadata);
+
+    const receipt = await network.actions.createSpace(auth.web3, {
+      controller: web3.value.account,
+      votingDelay: settings.votingDelay,
+      minVotingDuration: settings.minVotingDuration,
+      maxVotingDuration: settings.maxVotingDuration,
+      proposalThreshold: BigInt(settings.proposalThreshold),
+      qorum: BigInt(settings.quorum),
```
bonustrack, line 67: This PR works well, there is just these `qorum` variables that need to be changed

Sekhmet: Those were renamed in sx.js, but it was not released yet: https://github.com/snapshot-labs/sx.js/pull/133

I was thinking there would be new deployment so I could release new version with deployment change (and this will be included in that release) so we avoid two small releases, but I can do special release for that fix too.

bonustrack (approved):
tACK

PR #402, opened 2023-02-02
fix: overfetch spaces and proposals for better infinite loading

Sekhmet:
## Summary

Currently we have an issue if we have exactly `LIMIT` (or `n * LIMIT`) results in database. We will assume that there are more results to be fetched, but there might be none. Without server telling us (`hasMore` property), we can workaround it by fetching one result more and checking if there are more using this extra element. This causes more data to be transfered (could be solved by caching that extra result, but it might not be worth complexity), but provides better UX.

## Test plan

Fetching spaces/proposals works.

### DIFF
--- a/src/stores/proposals.ts
+++ b/src/stores/proposals.ts
@@ -58,11 +58,11 @@ export const useProposalsStore = defineStore('proposals', {
       record.value.loading = true;
 
       const proposals = await getNetwork(networkId).api.loadProposals(spaceId, {
-        limit: PROPOSALS_LIMIT
+        limit: PROPOSALS_LIMIT + 1
       });
 
-      record.value.proposals = proposals;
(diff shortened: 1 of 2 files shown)


bonustrack:
@Sekhmet Not too sure to understand the issue, why don't we just try load a new page and if there is no result we don't query it again?

Sekhmet:
@bonustrack It's UX thing:

**For spaces** we show Load more button if we think there are more spaces and we load those when user clisk the button, if we got `hasMoreSpaces` wrong user will click on it and nothing will actually happen, looking like something was broken
**For spaces and proposals** when we scroll down and we think there are more spaces it will show loading icon briefly, but nothing will be loaded (no new spaces/proposals).

It's minor thing, but I think it might be annoying (especially spaces with Load more button).

PR #403, opened 2023-02-02
build: add typecheck step

Sekhmet:
## Summary

Vite uses rollup/esbuild that don't actually typecheck the codebase, only transpile, meaning all type errors will go unnoticed. This adds new step with `vue-tsc` to run typechecks on builds and tests.


bonustrack (approved):
utACK

PR #404, opened 2023-02-02
feat: use networks explorer url in Settings

Sekhmet:
## Summary

This PR adds explorer URL from network for Space settigns page.

There is old `explorerUrl` still being used, but this will need some refactoring to improve network -> networkID mapping so leaving it out for now.


bonustrack (approved):
tACK

Issue #388, opened 2023-01-25
feat: suggest network change if user is logged on wrong network


bonustrack:
Currently if you try to cast a vote while you are on Ethereum mainnet it will just fail without showing any error on the UI. Instead we should suggest network change using: https://eips.ethereum.org/EIPS/eip-3326 https://docs.metamask.io/guide/rpc-api.html#wallet-switchethereumchain

PR #410, opened 2023-02-06
feat: request network change

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/388

This PR will prompt user to change network in Metamask if incorrect is selected.
ArgentX doesn't support network changes.

## Test plan

- Do actions that requires singing via Metamask while having wrong network selected.
- Metamask prompts you to change network.
- If rejected it won't do the action.
- If accepted action will continue executing.


bonustrack (approved):
tACK

Issue #408, opened 2023-02-06
feat: separate create space form in multiple steps


bonustrack:
We should separate the create space page into multiple steps, some of the section will require too much space on the UI that it wont fit well if we do all in a single page. The steps would be:

- Profile
- Network
- Strategie(s)
- Auth(s)
- Execution(s)
- Voting
- Controller

This issue depend on #407

PR #415, opened 2023-02-08
feat: add multi-step create space form

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/408
Depends on: https://github.com/snapshot-labs/sx-ui/pull/443

This PR adds multi-step form for creating spaces.

## Test plan

- Create space with new form

## Screenshots

https://user-images.githubusercontent.com/1968722/217544030-56a54fe4-2d26-4e78-84ab-266f8f2f884b.mp4


bonustrack:
On mobile I can not slide to the left side of the menu 
Screenshot_20230208_212144_Chrome

src/components/Block/SpaceFormProfile.vue
```diff
@@ -0,0 +1,53 @@
...
+      title: 'Website',
+      examples: ['Space website URL']
+    }
+  }
+};
+
+const formErrors = computed(() => {
+  const errors = validateForm(definition, props.form);
+
+  emit('errors', errors);
```
j0hnfl0w, line 42: I don't think it's a good idea to emit('errors', errors) on every form update even if we don't have any errors AND it looks like a side effect in computed property. So maybe we need to do it separately:
```
const formErrors = computed(() => validateForm(definition, props.form));
watch(formErrors, to => {
  if (Object.keys(to).length > 0) emit('errors', to);
});
```
We have this also in ```SpaceFormController.vue```

src/views/Create.vue
```diff
@@ -2,99 +2,73 @@
...
-      title: 'Proposal threshold',
-      examples: ['1']
-    },
-    quorum: {
-      type: 'string',
-      format: 'uint256',
-      title: 'Quorum',
-      examples: ['1']
-    }
+const pages = [
```
j0hnfl0w, line 14: So if it is a constant maybe better call it ```PAGES``` ?

src/views/Create.vue
```diff
@@ -153,60 +156,76 @@ watch(selectedNetworkId, () => {
...
+        />
+        <BlockSpaceFormStrategies
+          v-else-if="currentPage === 'executions'"
+          v-model="executionStrategies"
+          :available-strategies="selectedNetwork.constants.EDITOR_EXECUTION_STRATEGIES"
+          title="Execution strategies"
+          description="Lorem ipsum..."
+        />
+        <BlockSpaceFormVoting
+          v-if="currentPage === 'voting'"
```
j0hnfl0w, line 206: Why not use here ```v-else-if``` ?

src/views/Create.vue
```diff
@@ -153,60 +156,76 @@ watch(selectedNetworkId, () => {
...
+          :form="settingsForm"
+          @errors="v => handleErrors('voting', v)"
+        />
+        <BlockSpaceFormController
+          v-if="currentPage === 'controller'"
```
j0hnfl0w, line 211: Why not use here ```v-else-if``` ?

src/views/Create.vue
```diff
@@ -153,60 +156,76 @@ watch(selectedNetworkId, () => {
...
-          v-model="settingsForm"
-          :error="settingsFormErrors"
-          :definition="settingsDefinition"
+      <div class="mt-8 lg:mt-0">
+        <BlockSpaceFormProfile
+          v-if="currentPage === 'profile'"
+          :form="metadataForm"
+          @errors="v => handleErrors('profile', v)"
+        />
+        <BlockSpaceFormNetwork v-else-if="currentPage === 'network'" v-model="selectedNetworkId" />
```
j0hnfl0w, line 183: For consistency and readability we need every prop on new line here

src/views/Create.vue
```diff
@@ -153,60 +156,76 @@ watch(selectedNetworkId, () => {
...
+      <UiButton
+        v-if="showCreate"
+        class="w-full"
+        :loading="sending"
+        :disabled="submitDisabled"
+        @click="handleSubmit"
+      >
         Create
       </UiButton>
+      <UiButton v-else class="w-full" :disabled="nextDisabled" @click="handleNextClick">
```
j0hnfl0w, line 226: For consistency and readability we need every prop on new line here

j0hnfl0w (commented):
I couldn't get to the end of the form, its WIP I think. Found some things that can be improved.

src/views/Create.vue
Sekhmet: It's autoformatted by Prettier.

src/views/Create.vue
Sekhmet: It's autoformatted by Prettier.

src/components/Block/SpaceFormProfile.vue
Sekhmet: Agreed on separating those.

I think we need to emit, even if errors are empty though. Otherwise, if user corrects the form and erros are gone consumer will still think there are errors there.

Sekhmet:
> I couldn't get to the end of the form, its WIP I think. Found some things that can be improved.

You should be able to reach the end of the form. It won't let you past if you haven't filled in the form properly (through Next button or pills at the left).

j0hnfl0w:
ah, I see, I picked Ethereum (goerli) and there are no strategies there

Sekhmet:
@bonustrack updated, although I don't think it looks too good now. The idea was that the main content is centered and those navigation items are just addition. This way what you interact with is centered. Now it seems that everything is shifted to the right.

bonustrack:
@Sekhmet You think? It's not great but looks better to me. The design will need to be improved we can do that in another PR, we just don't have a design yet (it's actually what we use as challenge for hiring). I imagine a design where the page take the whole space (hiding the sidebar and header), something like this https://dribbble.com/shots/20530160-Minimal-sign-up-user-flow-Untitled-UI

bonustrack:
Some issue here can you reproduce?:
image

bonustrack (approved):
tACK

Issue #414, opened 2023-02-08
bug: sidebar spaces are not clickable


bonustrack:
It's not possible to click on a space from the sidebar on a mobile. This bug was introduced in #383

j0hnfl0w:
Hm, works good on iOS safari/chrome, a bit laggy in desktop device emulator.https://user-images.githubusercontent.com/97898172/217582720-e65d5c5e-3be4-45b2-80e4-0e8175be1d2a.MOV

j0hnfl0w:
Found this issue with Android chrome, dragging works fine but its hard to click

bonustrack:
@j0hnfl0w I'm able to drag the icons too but nothing happen when I click, also require 2 clicks to close the menu after trying to click on space. This is using Android 13 with Chrome 109.0.5414.118.

j0hnfl0w:
Yep, found a couple of issues with vuedraggable:
https://github.com/SortableJS/vue.draggable.next/issues/178
https://github.com/SortableJS/vue.draggable.next/issues/174
https://github.com/SortableJS/vue.draggable.next/issues/61
Will fix, now

PR #416, opened 2023-02-08
fix: click events on mobile in draggable

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/414

- fixing vuedraggable click events on mobile devices


bonustrack:
@j0hnfl0w It works for me but not all the time, sometime I get locked see: 
https://user-images.githubusercontent.com/16245250/217596937-e453c8c6-afc5-4c81-86c8-ba223af1af14.mp4

bonustrack:
Weird the same behavior is still happening for me

j0hnfl0w:
Checked again on my Android on different browsers:
https://www.loom.com/share/fb8030e2d2684ecda63c90000eac7d77

bonustrack:
Here is a record with latest version, for some reason when I click on a space that is already open it introduce some issue and next click failing, check: 

https://user-images.githubusercontent.com/16245250/220156924-85189e0d-cd45-4213-9494-bdd10004a7a8.mp4

Can you reproduce?

src/components/Sidebar.vue
```diff
@@ -10,9 +12,20 @@ const spacesStore = useSpacesStore();
...
+      :touch-start-threshold="35"
+      :item-key="i => i"
+      class="space-y-3 p-2"
+    >
       <template #item="{ element }">
-        <router-link :to="{ name: 'overview', params: { id: element } }" class="block">
+        <router-link
+          :to="{ name: 'overview', params: { id: element } }"
+          class="block"
+          @click="uiStore.sidebarOpen = false"
```
Sekhmet, line 27:
```suggestion
```

Do we need this change?

We do this:
https://github.com/snapshot-labs/sx-ui/blob/ecea212b7969c2467d43060fb6bd61680523aa51/src/App.vue#L39-L41

j0hnfl0w: We need to handle a click on already selected space, so route watcher doesn't handle that case.

Sekhmet (approved):
tACK

Issue #421, opened 2023-02-09
chore: move connectors icons into the repository


bonustrack:
Currently we are using IPFS to resolve images for connectors, but it take very long sometime to appear, we should instead move these images to the repository directly 
<img width="485" alt="image" src="https://user-images.githubusercontent.com/16245250/217681267-982750df-d6fd-41c4-8d85-b58de93b8c15.png">

PR #423, opened 2023-02-09
feat: add local aliases for connector icons

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/421

This adds aliases for IPFS URLs that point to local files. Using aliases allows to also overwrite `@snapshot-labs/lock` returned connectors that have IPFS link hardcoded upstream.

## Test plan

- Open Connect wallet modal.
- Shown images are served locally.


bonustrack:
@Sekhmet Not sure we need a mapping of IPFS hashes here, is it just so if we add a new connector and don't have an image we have the IPFS image as fallback? If thats the case we can just make sure this doesn't happen. This implementation also assume we wont change the IPFS hash on Lock.js but that might happen.

Sekhmet:
> is it just so if we add a new connector and don't have an image we have the IPFS image as fallback?

I've done it this way, because `getInjected` from `@snapshot-labs/lock` returns new connectors, that have their own IPFS hash defined upstream, so we can't just override them with local ones. We can force metamask from `getInjected` to get static icon (like in 1ed4894), but if there is connection that we don't have set up locally we can't override it (for example Exodus).

Alternative solution would be to define second mapping object (id -> icon), but I tried that and it's a mess because of `injected` being used for multiple different connectors (with different icons) so unless we are okay with showing Metamask icon for every injected provider it will be even more complex than what we have now (and had previously).

src/helpers/connectors.ts
```diff
@@ -0,0 +1,78 @@
...
+  walletconnect: {
+    id: 'walletconnect',
+    name: 'WalletConnect',
+    network: '1',
+    icon: walletconnectIcon,
+    options: {
+      rpc: {
+        '1': 'https://cloudflare-eth.com',
+        '4': 'https://eth-rinkeby.alchemyapi.io/v2/twReQE9Px03E-E_N_Fbb3OVF7YgHxoGq',
+        '42': 'https://eth-kovan.alchemyapi.io/v2/QCsM2iU0bQ49eGDmZ7-Y--Wpu0lVWXSO'
```
bonustrack, line 26: These networks are shutdown we can remove them
```suggestion
        '1': 'https://cloudflare-eth.com'
```

src/helpers/connectors.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+export function mapConnectorId(sourceName: string) {
+  if (sourceName === 'metamask') return 'injected';
+  if (sourceName === 'coinbase') return 'walletlink';
+
+  return sourceName;
+}
+
+export function getConnectorIconUrl(url) {
```
bonustrack, line 74: We could add our own function to replace getUrl from helper to transform urls like `ipfs://` with the defined `VITE_IPFS_GATEWAY` and also support `https://` urls.
https://github.com/snapshot-labs/sx-ui/blob/a64a801126173baf069dfe34f5ba153186f38869/src/helpers/utils.ts#L34-L36

bonustrack (approved):
tACK

Issue #422, opened 2023-02-09
feat: contacts page


bonustrack:
Add a new page "Contacts", that looks like this: 
<img width="1433" alt="image" src="https://user-images.githubusercontent.com/16245250/217686354-99d5e7ae-d92c-40a8-af64-98c6c5db769b.png">

This page is not specific to space. The page have a button +, when we click on it it open a modal that look like this:
<img width="505" alt="image" src="https://user-images.githubusercontent.com/16245250/217688322-fd580de4-4e59-4790-ae82-833b5c399208.png">

- The contact list should be stored on user localStorage.
- On mouse over a contact, a bin icon should appear in the right side of the line.
- The contact list should be proposed (along with connected account if available) in the list when we click on an input for address for example on the send token modal: 
<img width="514" alt="image" src="https://user-images.githubusercontent.com/16245250/217687147-a4ecd362-1cb2-4ebf-a1e9-6275abdd148b.png">

PR #424, opened 2023-02-09
feat: contacts

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/422

- added contacts page in /account/contacts, add/edit/delete contacts here
- contacts stores in localStorage
- contacts are available in in ContactPicker

### DIFF
--- a/src/components/Block/ContactPicker.vue
+++ b/src/components/Block/ContactPicker.vue
@@ -2,21 +2,7 @@
 import { computed } from 'vue';
 import { useAccount } from '@/composables/useAccount';
 import { shorten } from '@/helpers/utils';
-
-const CONTACTS = [
-  {
-    name: 'Sekhmet',
-    address: '0x556B14CbdA79A36dC33FcD461a04A5BCb5dC2A70'
-  },
-  {
-    name: 'Definitely not Sekhmet',
-    address: '0x537f1896541d28F4c70116EEa602b1B34Da95163'
-  },
-  {
-    name: 'WETH',
-    address: '0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6'
-  }
-];
+import { useContactsStore } from '@/stores/contacts';
 
 const props = defineProps<{
   searchValue: string;
@@ -28,16 +14,17 @@ const emit = defineEmits<{
 }>();
 
 const { account } = useAccount();
+const contactsStore = useContactsStore();
 
 const allContacts = computed(() => {
-  if (!account) return CONTACTS;
+  if (!account) return contactsStore.contacts;
 
   return [
     {
       name: 'You',
       address: account
     },
-    ...CONTACTS
+    ...contactsStore.contacts
   ];
 });
 
--- a/src/components/Modal/ContactEditor.vue
+++ b/src/components/Modal/ContactEditor.vue
@@ -1,0 +1,74 @@
+<script setup lang="ts">
+import { reactive, computed, watch } from 'vue';
+import { clone } from '@/helpers/utils';
+import { useContactsStore } from '@/stores/contacts';
+
+const DEFAULT_FORM_STATE = {
+  name: '',
+  address: ''
+};
+
+const props = defineProps<{
+  open: boolean;
+  initialState?: any;
+}>();
+
+const emit = defineEmits<{
+  (e: 'close');
+}>();
+
+const contactsStore = useContactsStore();
+
+const form: {
+  name: string;
+  address: string;
+} = reactive(clone(DEFAULT_FORM_STATE));
+
+const formValid = computed(() => form.name !== '' && form.address !== '');
+
+function handleSubmit() {
+  contactsStore.saveContact(clone(form));
+  emit('close');
+}
+
+watch(
+  () => props.open,
+  () => {
+    if (props.initialState) {
+      form.name = props.initialState.name;
+      form.address = props.initialState.address;
+    } else {
+      form.name = DEFAULT_FORM_STATE.name;
+      form.address = DEFAULT_FORM_STATE.address;
+    }
+  }
+);
+</script>
+
+<template>
+  <UiModal :open="open" @close="$emit('close')">
+    <template #header>
+      <h3 v-text="initialState?.address ? 'Edit contact' : 'Add contact'" />
+    </template>
+    <div class="s-box p-4">
+      <SIString
+        v-model="form.name"
+        :definition="{
+          type: 'string',
+          title: 'Name'
+        }"
+      />
+      <SIString
+        v-model="form.address"
+        :definition="{
+          type: 'string',
+          title: 'Address',
+          examples: ['Address or ENS']
+        }"
+      />
+    </div>
+    <template #footer>
+      <UiButton class="w-full" :disabled="!formValid" @click="handleSubmit"> Confirm </UiButton>
+    </template>
+  </UiModal>
+</template>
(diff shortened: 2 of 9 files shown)


src/components/Block/ContactPicker.vue
```diff
@@ -2,21 +2,7 @@
...
-  {
-    name: 'Definitely not Sekhmet',
-    address: '0x537f1896541d28F4c70116EEa602b1B34Da95163'
-  },
-  {
-    name: 'WETH',
-    address: '0xb4fbf271143f4fbf7b91a5ded31805e42b2208d6'
-  }
-];
+import { useContactsStore } from '@/stores/contacts';
```
bonustrack, line 5: We should hide the logged account if it's already available as contact to avoid duplicates.

src/views/Account/Contacts.vue
```diff
@@ -0,0 +1,72 @@
...
+      :key="contact.address"
+      class="px-4 py-3 border-b flex"
+      @mouseover="contactOver = contact.address"
+      @mouseleave="contactOver = null"
+    >
+      <div class="flex-auto flex items-center min-w-0" @click="onContactClick(contact)">
+        <Stamp :id="contact.address" type="token" :size="32" />
+        <div class="flex flex-col ml-3 leading-[22px] min-w-0 pr-2 md:pr-0">
+          <h4 class="text-skin-link" v-text="contact.name" />
+          <div class="text-sm truncate" v-text="contact.address" />
```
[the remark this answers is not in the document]
j0hnfl0w: But in ContactPicker.vue we have
```
<div class="text-skin-link" v-text="shorten(contact.name, 24)" />
<div class="text-sm text-ellipsis overflow-hidden" v-text="contact.address" />
```
Should it all be consistent or it just different cases?

bonustrack:
- Borders should stop at the red line
https://user-images.githubusercontent.com/16245250/218190017-58c85477-8920-4155-912d-d2c3b4091c20.png
Like this:
https://user-images.githubusercontent.com/16245250/218190185-debad32c-69a1-4dd1-9cfa-dd48a7a9a114.png
- Can you add a message if there isn't any contact similar than for proposals like this:
https://user-images.githubusercontent.com/16245250/218190324-7efd32a1-4f22-4e35-b7ca-b79ba5ebce25.png
But with the text "There are no contacts here." 
- There is a bug when I click on Space settings link it send me to "/settings" url with empty page
- Can you reduce the space between the button edit and delete to be the same as in the editor execution section?
- We shouldn't be able to add the same contact (currently it overwrite when adding duplicate this can be confusing for user). The address input can appear with red border if a duplicate is found. You can find an example of error input on the editor in the execution contract call section if you type a wrong address in the input

src/stores/contacts.ts
```diff
@@ -0,0 +1,25 @@
...
+          contact[key] = value;
+        });
+      } else {
+        this.contacts.unshift(payload);
+      }
+    },
+    deleteContact(address: string) {
+      this.contacts = this.contacts.filter(contact => contact.address !== address);
+    }
+  }
```
Sekhmet, line 24: I wonder, storing them as Record<address, Contact> would make handling those actions simpler. What do you think?

src/components/Block/ContactPicker.vue
```diff
@@ -28,16 +14,19 @@ const emit = defineEmits<{
 }>();
 
 const { account } = useAccount();
+const contactsStore = useContactsStore();
 
 const allContacts = computed(() => {
-  if (!account) return CONTACTS;
+  if (!account) return contactsStore.contacts;
+  if (contactsStore.contacts.find(contact => contact.address === account))
+    return contactsStore.contacts;
```
Sekhmet, line 22: nit. when `if` splits into two lines I'd say it makes sense to use `{}` so mistakes are less likely to happen if we ever need to add more statements in if body.

src/components/Nav.vue
```diff
@@ -1,28 +1,51 @@
...
   settings: {
-    name: 'Settings'
+    contacts: {
+      name: 'Contacts',
+      icon: IHUsers
+    }
   }
 };
+const navGroupKey = computed(() => String(route.matched[0]?.name));
+const navGroup = computed(() => navGroups[navGroupKey.value || '']);
```
Sekhmet, line 43: nit. I'd rename it to something more undersandable, maybe `NAVIGATION_CONFIG`, `currentRouteName`, `navigationItems`.

Sekhmet (commented):
If I try to edit existing contact, and only change contact name it won't allow me to save.

src/stores/contacts.ts
j0hnfl0w: Maybe it would be easier here, but then we need somehow to sort contacts, now we unshift new contact and its easy - new contacts are at the top. And if we will add some more properties to contact it would be easier )

bonustrack (approved):
tACK

bonustrack:
Can you fix conflicts @j0hnfl0w?

bonustrack:
There is an issue in explore page, we should fix it before merge

vite.config.ts
```diff
@@ -20,6 +20,7 @@ export default defineConfig({
   plugins: [
     vue({ reactivityTransform: true }),
     Components({
+      dts: true,
```
bonustrack, line 23: This can be removed if its the default

src/components/Nav.vue
```diff
@@ -31,16 +54,13 @@ const items = {
     <div class="h-[72px] border-b" />
     <div class="py-4">
       <router-link
-        v-for="(item, i) in items"
-        :key="i"
-        :to="{ name: i }"
+        v-for="(item, itemKey) in navigationItems"
```
bonustrack, line 57:
```suggestion
        v-for="(item, key) in navigationItems"
```

src/components/Nav.vue
```diff
@@ -1,28 +1,51 @@
-<script setup>
+<script lang="ts" setup>
```
bonustrack, line 1: Let's make another component for the settings nav

src/components/Modal/ContactEditor.vue
```diff
@@ -0,0 +1,88 @@
...
+      />
+      <SIString
+        v-model="form.address"
+        :error="addressDuplicated"
+        :definition="{
+          type: 'string',
+          title: 'Address',
+          examples: ['Address or ENS']
+        }"
+      />
```
Sekhmet, line 82: Maybe we should use SIObject there, where we would run normal formValidation.
```js
const formErrors = computed(() => validateForm(definition, form));
```

But to make addressDuplicated work we combine those two validations
```js
const errors = validateForm(definition, form)
if (addressDuplicated) baseErrors.address = 'Address duplicated'
return errors
```

j0hnfl0w: Maybe we can keep it simple until we extend contact form to more than 2 fields?

Sekhmet: I think that would make it simpler, because it would reuse already existing `validateForm` that we use across the app already.

src/components/Modal/EditContact.vue
```diff
@@ -0,0 +1,97 @@
...
+
+const addressDuplicated = computed(() => {
+  const duplicate = contactsStore.contacts.find(c => c.address === form.address);
+  if (duplicate) {
+    if (duplicate.address === props.initialState?.address) {
+      return false;
+    }
+    return true;
+  }
+  return false;
```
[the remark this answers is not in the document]
j0hnfl0w: Agreed, its better to return early

Sekhmet (approved):
tACK

Issue #417, opened 2023-02-08
fix: show edit profile button only for space controller


bonustrack:
The button to edit space profile should only be visible if you are logged with the address controller of the space
<img width="1464" alt="image" src="https://user-images.githubusercontent.com/16245250/217672852-166e1e1c-4e4a-491e-94be-75f58370c03f.png">

PR #426, opened 2023-02-09
fix: space is editable only for controller

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/417

- only logged in user and a controller can edit space


bonustrack:
Haven't been able to test there is a bug not related to this PR, it seem that we store controller address in the API with there is an extra 0:
image

bonustrack (approved):
tACK

src/views/Space/Overview.vue
```diff
@@ -20,6 +22,7 @@ onMounted(() => {
 const spaceIdComposite = `${props.space.network}:${props.space.id}`;
 
 const spaceStarred = computed(() => spacesStore.starredSpacesIds.includes(spaceIdComposite));
+const spaceIsEditable = computed(() => props.space.controller === web3.value.account);
```
Sekhmet, line 25:
```suggestion
const spaceIsEditable = computed(
  () => props.space.controller.toLowerCase() === web3.value.account.toLowerCase()
);
```

For Ethereum we might need to convert it to lowercase. I think that overall we should have some kind of helper for comparing addresses (that ignores 0-pads, casing etc.), normalizing it on server-side is okay, but I imagine there may be more bugs like this because some new source uses different format (let's say if ArgentX starts padding addresses).
What do you think @bonustrack?

bonustrack: Yes I think it makes sense to use lowercase to check if you are owner on the UI, some wallet may expose a lowercased address. I would just enforce the right casing / padding on the API.

Sekhmet (approved):
tACK

Issue #418, opened 2023-02-08
style: option for sticky label bars


bonustrack:
We should make an option for the Label component to be sticky, the component is this one:
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/217673528-aa725c80-a4a6-473c-9c75-137b86a7ea1e.png">

The label bar should scroll as usual, until it goes to this point:
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/217674157-f5fafb75-18b3-4331-adb3-18a7e45f5903.png">

Then it should stick at the same level with similar behavior than the blue bar here:
https://www.kindacode.com/wp-content/uploads/2022/07/Bootstrap-5-sticky-navbar.gif

Ideally on the page where there is multiple label we would make it so only one label bar is sticky at once.

PR #428, opened 2023-02-10
feat: add sticky prop to Label

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/418

- added sticky prop, without default state
- removed overflow-hidden from App wrapper div, cos sticky wont work with it [source](https://www.designcise.com/web/tutorial/how-to-fix-issues-with-css-position-sticky-not-working)

### DIFF
--- a/src/components/Label.vue
+++ b/src/components/Label.vue
@@ -9,7 +9,7 @@ defineProps({
   <h4
     class="eyebrow border-b py-2 px-4 text-skin-text"
     :class="{
-      'sticky top-[71px] bg-skin-bg': sticky
+      'sticky top-[72px] bg-skin-bg': sticky
     }"
     v-text="label"
   />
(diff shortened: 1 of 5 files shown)


src/components/Label.vue
```diff
@@ -1,7 +1,16 @@
...
+  sticky: Boolean
+});
 </script>
 
 <template>
-  <h4 class="eyebrow border-b py-2 px-4 text-skin-text" v-text="label" />
+  <h4
+    class="eyebrow border-b py-2 px-4 text-skin-text"
+    :class="{
+      'sticky top-[71px] bg-skin-bg': sticky
```
bonustrack, line 12: I think it should be 72px here, otherwise if you are on the settings page the "Voting" tab will move from 1px when you scroll
```suggestion
      'sticky top-[72px] bg-skin-bg': sticky
```

bonustrack (approved):
tACK

Issue #419, opened 2023-02-08
feat: avatar upload on edit profile modal


bonustrack:
We should add a way to upload an avatar from the settings page, the form would look like this:
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/217672782-25a7b32b-9767-44b3-8634-a0def973a5fc.png">

In the modal the height of the mini cover is 100px and the size of the avatar 80px.

On mouse over the avatar, we should make the image with 80% opacity and add an icon "upload" in the center. We can use pineapple to upload the image, similar than what we did in snapshot.org here is the code: https://github.com/snapshot-labs/snapshot/blob/53472750345dc269701602bd49f109ab272b771c/src/composables/useImageUpload.ts

This will require some change in Stamp https://github.com/snapshot-labs/stamp to display images from SX.

PR #429, opened 2023-02-10
feat: upload space avatar wip

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/419

- [x] uploads avatar to ipfs with pineapple
- [ ] save avatar url to space.avatar, need to be an owner of the space => need to create my space
- [ ] add sx resolver to stamp.fyi service to serve sx space avatars
       like a space resolver [here](https://github.com/snapshot-labs/stamp/blob/master/src/resolvers/space.ts)
- [ ] upload avatar on space creation


bonustrack:
- We should add border radius on the avatar preview when adding a new avatar
- We should not show the input "avatar", this should be hidden
- Edit icon should be white, and appear only on mouse over
- Spacing should be smaller here:
image

Issue #338, opened 2022-10-15
feat: add Histoire


bonustrack:
https://histoire.dev/guide/vue3/getting-started.html

Dhaiwat10:
if you guys are accepting contributions from outside, would love to help out with this one!

bonustrack:
Awesome, go for it @Dhaiwat10 :)

j0hnfl0w:
@bonustrack there are still problems with tailwind and histoire, tried almost the same solution as in https://github.com/snapshot-labs/sx-ui/pull/341 but it wont work.
Then I tried to setup histoire with https://github.com/unocss/unocss and it works good, also migrated almost all tailwind config, so css experience and api are the same.

Sekhmet:
It has been integrated.

PR #430, opened 2023-02-10
refactor: try to setup histoire with unocss

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/338

For now histoire doesn't work good with tailwind, so here is a try to use it with unocss.

Need to polyfill all the tailwind api to avoid ui glitches...


bonustrack:
I'm closing this PR, it's unlikely we would move to unocss we can look again adding Histoire when issue is solved with Tailwind

Issue #435, opened 2023-02-13
fix: separate data from lists in spaces/proposals


Sekhmet:
Currently we have an issue if fetching space that is not in top 10 spaces, for example:
https://testnet.snapshotx.xyz/#/sn-tn2:0x042335d5d732075a702da38191d6b6f916acc84dfcbab0d73b64065856f3faaf

It will be fetched initially, but request to fetch 10 spaces for homepage will overwrite the whole list and it will be stuck at loading.

We should store data separately from any list, so they can be independent:
```js
{
  spaces: ['id1', 'id2', 'id3'],
  spacesMap: {
    'id1': {...},
    'id2': {...},
    'id3': {...},
  }
}
```

PR #436, opened 2023-02-13
fix: remove fetching spaces in Sidebar

Sekhmet:
It's no longer needed as spaces in sidebar now come from storage only.

This will fix issue with https://github.com/snapshot-labs/sx-ui/issues/435, but proper fix is still needed.


bonustrack (approved):
tACK

PR #437, opened 2023-02-14
feat: refactor routes to handle 404 errors

j0hnfl0w:
- like in [snapshot](https://github.com/snapshot-labs/snapshot/blob/develop/src/router.ts) we need to handle 404 error
- we can validate our composite space ID = [networkId, spaceId] with routerGuard validateIdGuard
- we can group all space related routes to space route children
- minor fixes in Proposal view and Proposal component


src/router.ts
```diff
@@ -10,8 +10,25 @@ import Proposal from '@/views/Proposal.vue';
...
+import { isAddress } from '@ethersproject/address';
+import { getNetwork } from '@/networks';
+
+const validateIdGuard = (to, from, next) => {
+  try {
+    const id = to.params.id;
+    const [networkId, spaceId] = id.split(':');
+    console.log(':beforeEnter,', networkId, spaceId);
+    getNetwork(networkId);
+    if (!isAddress(spaceId)) return next('/');
```
Sekhmet, line 22: This won't work for starknet spaces.

src/router.ts
```diff
@@ -10,8 +10,25 @@ import Proposal from '@/views/Proposal.vue';
...
 import Create from '@/views/Create.vue';
 
+import { isAddress } from '@ethersproject/address';
+import { getNetwork } from '@/networks';
+
+const validateIdGuard = (to, from, next) => {
+  try {
+    const id = to.params.id;
+    const [networkId, spaceId] = id.split(':');
+    console.log(':beforeEnter,', networkId, spaceId);
```
Sekhmet, line 20:
```suggestion
```

Sekhmet (changes requested):

src/router.ts
j0hnfl0w: So we don't need this validation then

j0hnfl0w:
So now this PR adds only grouping Space routes in children and 404 redirect to /

src/router.ts
Sekhmet: Maybe we do, we could validate both networks if we wanted. I don't know what we want, it was your idea I think :D

j0hnfl0w: I think we can add it later for both networks

src/router.ts
```diff
@@ -21,18 +23,21 @@ const routes: any[] = [
...
+    path: '/profile/:id',
+    name: 'user',
+    component: User
   },
-  { path: '/:space/proposal/:id?', name: 'proposal', component: Proposal },
-  { path: '/profile/:id', name: 'user', component: User },
-  { path: '/create', name: 'create', component: Create },
-  { path: '/explore', name: 'explore', component: Explore }
+  {
+    path: '/:pathMatch(.*)*',
```
Sekhmet, line 37: Maybe we can whitelist `:id` in `space` route, so it doesn't math anything that doesn't look like space ID and swallow those.
```
path: '/:id([^:]+:[A-Za-z0-9]+)',
```

Sekhmet: @j0hnfl0w what do you think?

Sekhmet (approved):
tACK

PR #438, opened 2023-02-14
refactor: separate data from lists in spaces/proposals

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/435

This PR separates data stored in store into two properties:
- `entitiesIdsList` - array of IDs of entities list on UI (list of spaces, list of proposals)
- `entities` - map of IDs -> Entity that is used when rendering lists of rendering single entity (outside of a list).

This solves few issues:
- Potential race conditions when fetching results and single entity at the same time (eg. proposal might be added twice in the list if we fetch single proposal and proposal list).
- Already fetched single entity might disappear if we fetch list later.

### DIFF
--- a/src/stores/spaces.ts
+++ b/src/stores/spaces.ts
@@ -73,13 +73,14 @@ export const useSpacesStore = defineStore('spaces', {
 
       this.networksMap = Object.fromEntries(
         results.map(result => {
+          const spacesIds = result.spaces.map(space => space.id);
+
           return [
             result.id,
             {
-              spacesIdsList: [
-                ...this.networksMap[result.id].spacesIdsList,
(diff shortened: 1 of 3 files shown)


bonustrack (approved):
tACK

src/stores/spaces.ts
```diff
@@ -68,9 +76,14 @@ export const useSpacesStore = defineStore('spaces', {
           return [
             result.id,
             {
-              spaces: overwrite
-                ? result.spaces
-                : [...this.networksMap[result.id].spaces, ...result.spaces],
+              spacesIdsList: [
```
Sekhmet, line 79: Needs overwrite option again.

Issue #432, opened 2023-02-13
bug: not able to create a proposal on a space with vanilla voting strategy


bonustrack:
For some reason on this space I'm not able to create a proposal:
https://testnet.snapshotx.xyz/#/sn-tn2:0x022a6667315cc87d7e92f994ac6b621a7a86f76f78442cf90155ae5a7fc27dc9/settings

Mana return this error:
```
{"jsonrpc":"2.0","error":{"code":500,"message":"unauthorized","data":{"errorCode":"StarknetErrorCode.TRANSACTION_FAILED"}},"id":null}
```

PR #439, opened 2023-02-14
feat: support vanilla executor

Sekhmet:
## Summary

We currently allow user to use vanilla executor in space alone, but we hardcode zodiac executor when creating proposal.
This PR adds support for vanilla as well.

Closes: https://github.com/snapshot-labs/sx-ui/issues/432

## Test plan

- Create proposal on http://localhost:8080/#/sn-tn2:0x022a6667315cc87d7e92f994ac6b621a7a86f76f78442cf90155ae5a7fc27dc9


bonustrack (approved):
tACK

PR #441, opened 2023-02-16
feat: move space list to explore page

bonustrack:
This PR move the list of space to a new page "Explore". This also remove the pagination for spaces to load all of them at once, this avoid the issue of item count being inconsistent cuz we loading data from 2 different sources.

image


Sekhmet (approved):
tACK

> this avoid the issue of item count being inconsistent cuz we loading data from 2 different sources.

what item count do you mean? do we display it anywhere?

bonustrack:
@Sekhmet I mean we cant predict how many spaces to return when querying from 2 sources. In the layout with 3 columns I need to have strict number of space to return on each page otherwise it create empty area, e.g. I want get 12 spaces but get 11.

PR #446, opened 2023-02-20
fix: only go back on propose/deploy if succeeded

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/442

## Test plan
- Without logging in go to proposal editor.
- Hit publish - it doesn't take you back.
- Submit proposal after logging in - it takes you back.


bonustrack (approved):
tACK

PR #448, opened 2023-02-21
fix: prevent iframe

### LINKED ISSUE #411: chore: avoid iframe with X-FRAME-OPTIONS
Allowing iframe my lead to clickjacking, we should enforce no iframe using X-FRAME-OPTIONS. The tricky part is that X-FRAME-OPTIONS should be set on server side but our site is static, so we may need helo from our hosting provider or find another way to enforce no iframe.

bonustrack:
Fixes #411 

This PR prevent usage of SX within an iframe, this is to avoid clickjacking.


Sekhmet (approved):
tACK

Issue #434, opened 2023-02-13
feat: display total support for each choices


bonustrack:
Currently on a proposal we show only progress bars for the scores, we should show the amount of voting power in support for each choice "For" "Against" and "Abstain".
**Design TBD**

bonustrack:
Can be like this for now:
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/221014618-4fe02f64-850b-4883-9f39-60536e77c8db.png">

Sekhmet:
@bonustrack what if one choice has low support and its total support label won't fit?


![Image](https://user-images.githubusercontent.com/1968722/221811718-f46c8c02-c978-4e06-b58d-fa3218f8dabb.png)


There is also case of quorum: if quorum is 1M and currently there are 1% votes For then For bar will be 1% of the total width, not enough to fit any text.

bonustrack:
@Sekhmet You could show only the percentage and vp for the choice with most support and then show the others at the same place on mouse over like this:

<img width="806" alt="image" src="https://user-images.githubusercontent.com/16245250/221873392-fed0274c-a5fc-4bc0-a39a-367a91066186.png">

<img width="819" alt="image" src="https://user-images.githubusercontent.com/16245250/221873635-2d3cc698-b179-4d3a-abba-814d7bebaa2a.png">

Sekhmet:
@bonustrack is it going to work like this in all cases, or only when other options won't fit otherwise?

bonustrack:
For all case @Sekhmet

PR #456, opened 2023-02-28
feat: display voting scores

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/434

Display voting scores in UI.

Changes:
- Sort results bar based on most votes (if there are more votes against, red bar will be first, otherwise showing green bar and red text is super weird).
- You can click (instead of hover, to make mobile UX better) to get all results displayed.

## Test plan

http://localhost:8080/#/sn-tn2:0x07e6e9047eb910f84f7e3b86cea7b1d7779c109c970a39b54379c1f4fa395b28/proposal/56

### DIFF
--- a/src/components/Results.vue
+++ b/src/components/Results.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { computed } from 'vue';
+import { ref, computed } from 'vue';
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

PR #461, opened 2023-03-03
fix(evm): use regular client for ethTx/vanilla authenticator

Sekhmet:
## Summary

Forgot that routing all votes/proposals through mana would mean that those that use ethTx or vanilla authenticator would make Mana's account proposer/voter in all of them, this also made it impossible for multiple people to vote for proposal, because all other person would seemingly already voted.

## Test plan

- Propose here: http://localhost:8080/#/gor:0xb3782e21cedaf9183da9c472f810cc2fd337c6a2
- Signature request shows up
- Propose here: http://localhost:8080/#/gor:0xd381e6be943e254ab992cbc8698fc95b215109ea
- Transaction request shows up

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -28,7 +28,7 @@ export function useActions() {
     console.log('envelope', envelope);
 
     // TODO: unify send/soc to both return txHash under same property
-    if (network.hasRelayer) {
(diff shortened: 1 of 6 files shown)


bonustrack (approved):
tACK

PR #462, opened 2023-03-03
fix(evm): only convert Date.now() timestamp for VP calculation

Sekhmet:
## Summary

We should only convert `Date.now()`, `snapshot` timestamp is already in proper format.

## Test plan

- Go to proposal before you had delegated funds in space with CompToken strategy, you have 0 VP.
- Go to proposal after you had funds delegated, you have more VP.

## Screenshots

https://user-images.githubusercontent.com/1968722/222683649-b6cb99b4-9f92-4cd4-b8b1-71308f095218.mp4


bonustrack (approved):
utACK

PR #464, opened 2023-03-03
fix: do not add ago to now time

Sekhmet:
## Summary

Changes:
- use `fromNow` instead of `toNow` to make it more natural (right now `future` uses `ago` and `past` uses `left`).
- add custom cases for now: for future use `in few seconds`, for past `just now`.


bonustrack (approved):
tACK

PR #470, opened 2023-03-03
chore: remove snapshot-oc package from config

Sekhmet:
We no longer use this package (it was merged into sx.js)


j0hnfl0w (approved):
utACK

Issue #453, opened 2023-02-23
fix: remove Vanilla authenticator


bonustrack:
We should remove support for Vanilla authenticator in the UI and SX.js. This can lead to misleading activities in which anyone can be anyone.

PR #472, opened 2023-03-03
feat(evm): remove vanilla authenticator support

Sekhmet:
## Summary

Closes https://github.com/snapshot-labs/sx-ui/issues/453

This PR disabled vanilla authenticator from supported list (no longer possible to vote and propose with it) and editor (no longer possible to create space with it). Still left it in UI, so in old spaces it will show up as Vanilla.


j0hnfl0w (approved):
utACK

Issue #468, opened 2023-03-03
feat: remove quorum from Space create form for EVM spaces


Sekhmet:
Quorum for EVM spaces is defined on the execution strategy, not on space. We should hide it in Create space form for EVM (but keep it for Starknet for now).

PR #474, opened 2023-03-06
feat: remove quorum for evm

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/468


src/components/Block/SpaceFormVoting.vue
```diff
@@ -43,12 +46,14 @@ const definition = {
       title: 'Proposal threshold',
       examples: ['1']
     },
-    quorum: {
-      type: 'string',
-      format: 'uint256',
-      title: 'Quorum',
-      examples: ['1']
-    }
+    ...(!evmNetworks.includes(props.selectedNetworkId) && {
```
Sekhmet, line 49: We need to change `required` too.

Sekhmet (changes requested):
image

It's impossible to process in this form:
image

j0hnfl0w:
hm, all of it dynamic quorum changes looks ugly, maybe there is a better way. Maybe it can be only temporary solution.

Sekhmet:
> hm, all of it dynamic quorum changes looks ugly, maybe there is a better way. Maybe it can be only temporary solution.

Maybe we can avoid those ugly changes by enabling additionalProperties on the definition so we can just pass quorum in without worrying about mapping types/data.

j0hnfl0w:
yep, here will be the same solution

j0hnfl0w:
@Sekhmet looked again and decided to leave this dynamic staff, we still need to hide quorum from definition

Sekhmet (changes requested):

src/views/Create.vue
```diff
@@ -81,7 +81,9 @@ const settingsForm: SpaceSettings = reactive(
     minVotingDuration: 0,
     maxVotingDuration: 86400,
     proposalThreshold: '1',
-    quorum: '1'
+    ...(!evmNetworks.includes(selectedNetworkId.value) && {
+      quorum: '1'
+    })
```
Sekhmet, line 86: This doesn't really work anyway (https://github.com/snapshot-labs/sx-ui/pull/474/files#r1133918323), so let's revert it.

src/views/Create.vue
```diff
@@ -4,7 +4,7 @@ import { useRouter } from 'vue-router';
 import { useActions } from '@/composables/useActions';
 import { useWeb3 } from '@/composables/useWeb3';
 import { clone } from '@/helpers/utils';
-import { getNetwork } from '@/networks';
+import { getNetwork, evmNetworks } from '@/networks';
```
Sekhmet, line 7: Dead import.

Sekhmet (approved):
tACK

Issue #480, opened 2023-03-06
feat: make it possible to select "No treasury" in Space form


Sekhmet:
It should be possible to select No treasury in Space form (create and edit):
![image](https://user-images.githubusercontent.com/1968722/223200248-801e9e23-68d6-4199-9be2-ca9da7e92abb.png)

j0hnfl0w:
@Sekhmet so walletNetwork can be null or something like "none"?

Sekhmet:
I think ideally it would be null.

PR #486, opened 2023-03-07
feat: no-treasury in space create form

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/480

### Conditions
- if walletNetwork === null we hide walletAddress input
- if walletNetwork !== null we must fill walletAddress

### DIFF
--- a/src/components/Block/SpaceFormProfile.vue
+++ b/src/components/Block/SpaceFormProfile.vue
@@ -70,13 +70,21 @@ const definition = computed(() => {
         title: 'Treasury network',
         nullable: true
       },
-      walletAddress: {
-        type: props.form.walletNetwork === null ? 'null' : 'string',
-        format: 'address',
-        title: 'Treasury address',
-        examples: ['0x0000…'],
-        hidden: props.form.walletNetwork === null
-      }
+      walletAddress:
+        props.form.walletNetwork === null
+          ? {
+              type: 'null',
+              title: 'Treasury address',
+              examples: ['0x0000…']
+            }
+          : {
+              type: 'string',
+              title: 'Treasury address',
+              examples: ['0x0000…'],
+              format: 'address',
+              minLength: 1,
+              hidden: props.form.walletNetwork === null
+            }
     }
   };
 });
--- a/src/components/Block/SpaceFormProfile.vue
+++ b/src/components/Block/SpaceFormProfile.vue
@@ -16,6 +16,7 @@ const props = withDefaults(
 const emit = defineEmits<{
   (e: 'errors', value: any);
   (e: 'pick', field: any);
+  (e: 'no-network');
 }>();
 
 const availableNetworks = enabledNetworks.map(id => ({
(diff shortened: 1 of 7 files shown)


src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,79 @@ const props = withDefaults(
...
+  };
+});
+
+watch(
+  () => props.form.walletNetwork,
+  to => {
+    if (to === null) emit('no-network');
   }
-};
+);
```
Sekhmet, line 89: Why do we need it? Wouldn't normal form updates handle it already?

src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,79 @@ const props = withDefaults(
...
+      },
+      discord: {
+        type: 'string',
+        title: 'Discord',
+        examples: ['Discord handle']
+      },
+      walletNetwork: {
+        type: ['string', 'null'],
+        enum: [...enabledNetworks, null],
+        options: [...availableNetworks, { id: null, name: 'No treasury' }],
```
Sekhmet, line 69: Can we put `No treasury` first?

src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,79 @@ const props = withDefaults(
...
+        title: 'Treasury network',
+        nullable: true
+      },
+      walletAddress: {
+        type: props.form.walletNetwork === null ? 'null' : 'string',
+        format: 'address',
+        title: 'Treasury address',
+        examples: ['0x0000…'],
+        hidden: props.form.walletNetwork === null
```
Sekhmet, line 78: I wonder if it would make sense to just remove `hidden` and check for `type === null`.

src/components/Block/SpaceFormProfile.vue
j0hnfl0w: When we set walletNetwork to null we need to set walletAddress to null to, and it can be made outside SpaceFormProfile.vue

src/components/Block/SpaceFormProfile.vue
j0hnfl0w: I think it is more explicitly and we can use it somewhere else when we need to hide form fields based on some condition.

src/components/Block/SpaceFormProfile.vue
Sekhmet: Makes sense, can we handle it internally though, so it works out of the box?

For example now it won't work in edit space modal.

j0hnfl0w: Hm, it is not an option to mutate props.form, so need another way

Sekhmet: @j0hnfl0w good point, let's keep it as-is then, but fix edit space modal to also handle this event.

src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,79 @@ const props = withDefaults(
...
+      },
+      walletNetwork: {
+        type: ['string', 'null'],
+        enum: [null, ...enabledNetworks],
+        options: [{ id: null, name: 'No treasury' }, ...availableNetworks],
+        title: 'Treasury network',
+        nullable: true
+      },
+      walletAddress: {
+        type: ['string', 'null'],
```
Sekhmet, line 74: I think type there should be conditional, otherwise it's possible to select network, without putting in address and it's valid.

Sekhmet (commented):
Other than that it looks good to me.

src/components/Block/SpaceFormProfile.vue
j0hnfl0w: So we need to change initial walletAddress value to null then

Sekhmet (commented):
Found two more issues:
1. if I select network, type something in walletAddress and delete it all it still accepts it as correct address.
2. There are some warnings in the console.
image

src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,87 @@ const props = withDefaults(
...
+              examples: ['0x0000…']
+            }
+          : {
+              type: 'string',
+              title: 'Treasury address',
+              examples: ['0x0000…'],
+              format: 'address',
+              minLength: 1,
+              hidden: props.form.walletNetwork === null
+            }
```
Sekhmet, line 87: At this point I'm thinking that maybe we should just add `walletAddress` conditionally (it either is there, or it's not), and we can just toggle `additionalProperties` to true. With that we wouldn't need `hidden`, `type: 'null'` and all those extra changes - with benefit of being simpler and not adding definitions that basically mean: "nothing".

src/components/Block/SpaceFormProfile.vue
```diff
@@ -16,67 +16,87 @@ const props = withDefaults(
...
+              title: 'Treasury address',
+              examples: ['0x0000…']
+            }
+          : {
+              type: 'string',
+              title: 'Treasury address',
+              examples: ['0x0000…'],
+              format: 'address',
+              minLength: 1,
+              hidden: props.form.walletNetwork === null
```
Sekhmet, line 86:
```suggestion
```

We don't need hidden now because it's always false there.

src/components/S/IObject.vue
```diff
@@ -44,8 +44,19 @@ const inputValue = computed({
...
 
-const getComponent = (property: { type: string; format: string; enum?: string[] }) => {
-  switch (property.type) {
+const getComponent = (property: {
+  type: string;
+  format: string;
+  enum?: string[];
+  hidden?: boolean;
+}) => {
+  if (property.hidden) return null;
```
Sekhmet, line 53: We can revert this too.

Sekhmet (commented):
Other than that it looks good to me (again :smile: )

src/components/S/IObject.vue
```diff
@@ -44,8 +44,17 @@ const inputValue = computed({
   }
 });
 
-const getComponent = (property: { type: string; format: string; enum?: string[] }) => {
-  switch (property.type) {
+const getComponent = (property: {
+  type: string;
+  format: string;
+  enum?: string[];
+  hidden?: boolean;
```
Sekhmet, line 51:
```suggestion
```

Sekhmet (approved):
tACK

Issue #477, opened 2023-03-06
feat: display errors on UI


Sekhmet:
Some things fail (voting, proposing) but there is no feedback displayed to the user.

PR #496, opened 2023-03-10
feat: add generic UI errors for actions

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/477

This PR adds generic UI errors to actions. They are the same for every action, but should be okay for now, at least you see when something failed.

## Test plan
- Add throw in some action in `useActions`.
- Try this action.
- Generic message appears.

### DIFF
--- a/src/stores/ui.ts
+++ b/src/stores/ui.ts
@@ -1,7 +1,13 @@
 import { defineStore } from 'pinia';
 import { getNetwork } from '@/networks';
 import { lsSet, lsGet } from '@/helpers/utils';
-import { NetworkID } from '@/types';
+import { NotificationType, NetworkID } from '@/types';
+
+type Notification = {
+  id: string;
+  type: NotificationType;
+  message: string;
+};
 
 type PendingTransaction = {
   networkId: NetworkID;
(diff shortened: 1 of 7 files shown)


src/stores/ui.ts
```diff
@@ -17,12 +23,27 @@ function updateStorage(pendingTransactions: PendingTransaction[]) {
...
+    addNotification(type: NotificationType, message: string, timeout = 5000) {
+      const id = crypto.randomUUID();
+
+      this.notifications.push({
+        id,
+        type,
+        message
+      });
+
+      setTimeout(() => this.dismissNotification(id), 5000);
```
j0hnfl0w, line 42: We have timeout default function param = 5000, lets use it here.

j0hnfl0w (commented):
We will handle only errors in useActions? We can also handle errors from proposalsStore.fetchProposal() for example.

j0hnfl0w (approved):

Sekhmet:
> We will handle only errors in useActions? We can also handle errors from proposalsStore.fetchProposal() for example.

I'm not sure if such alerts would work for fetching data, as we do a lot of fetching and it might be hard to understand what is the alert that you received relating to. I think for fetching data we should have extra state on UI, where we display "Proposal was not found" instead of proposal (loading/loaded/failed).

bonustrack (approved):
utACK

PR #499, opened 2023-03-13
fix: mark withDetails as optional

Sekhmet:
It's actually optional, but type was set as required, causing it to throw warnings.


j0hnfl0w (approved):
tACK

Issue #498, opened 2023-03-13
bug: hide buttons to cast vote within voting delay period


bonustrack:
On a space with voting delay example: https://snapshotx.xyz/#/gor:0xff0a04a1e441903f3f5631cfaf6c024cd923a47f we should not show button to cast a vote within the voting delay period.

PR #500, opened 2023-03-13
feat: prevent voting before voting delay ended

Sekhmet:
We should prevent voting during voting delay period.

Closes: https://github.com/snapshot-labs/sx-ui/issues/498

## Test plan

- Create proposal on http://localhost:8080/#/gor:0xff0a04a1e441903f3f5631cfaf6c024cd923a47f
- You can't vote on it right away.


bonustrack (approved):
tACK

PR #504, opened 2023-03-15
fix: keep balances/nfts state inside of composable

Sekhmet:
Currently NFTs and balances are global, meaning that switching from one space to another you will see old results.

## Test plan
- Go to one space (Decimals 16 + 18) -> Treasury -> NFTs.
- **Without refreshing**.
- Go to different space (Or-Land) -> Treasury -> NFTs.
- NFTs and balances were loading initially, you didn't see values from different space.


j0hnfl0w (approved):
utACK

Issue #492, opened 2023-03-09
style: space menu should be hidden on proposal and editor pages


bonustrack:
We shouldn't see the space menu on proposal and editor pages
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/223934340-15354216-ff2c-4f72-80c4-3eef91a1c7c9.png">

j0hnfl0w:
So if we hide menu on proposal page we need somehow to return to space page
![image](https://user-images.githubusercontent.com/97898172/225908290-85255b3b-a887-4947-9425-18d4ef40e7eb.png)
Just like we can return from editor with btn
![image](https://user-images.githubusercontent.com/97898172/225908481-8469b841-bbda-46c5-a1c6-ee5119464401.png)

j0hnfl0w:
@bonustrack ping

Sekhmet:
@j0hnfl0w it was like this before, this menu being visible is regression from your PR I believe that refactored how routes are  defined, you can check how it looked like before it.

Sekhmet:
I tested latest #508 and found following issues:

Extra bar shows up on explore page:
![image](https://user-images.githubusercontent.com/1968722/232809884-d967722e-6876-488f-baa4-ff26d9b62e6b.png)

Search proposals shows up on proposal page (actually same as master, because I used the same if check as we had for sidebar).
![image](https://user-images.githubusercontent.com/1968722/232810558-19b93d1c-3b83-4cbf-9a96-c49a68562fb7.png)

This tends to have something broken all the time, what do you think of making it a separate PR? It will be easier to review and test if it it's not part of 20 file PR already.

PR #508, opened 2023-03-22
feat: nested navigation

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/492


bonustrack:
The spaces list should appear with component than on Explore page with a layout that has same width

j0hnfl0w:
@Sekhmet we got overlapping logic in spacesStore and useSpaces composable now. spacesStore is made to use only one spaces list so I moved logic to composable, so maybe we can use the same composable from store?

src/router.ts
```diff
@@ -40,7 +41,8 @@ const routes: any[] = [
     name: 'settings',
     component: Settings,
     children: [
-      { path: '', redirect: '/settings/contacts' },
+      // { path: '', redirect: '/settings/contacts' },
```
Sekhmet, line 44: Either remove it entirely, or add comment why it's commented out.

src/networks/types.ts
```diff
@@ -3,7 +3,7 @@ import type { Signer } from '@ethersproject/abstract-signer';
 import type { MetaTransaction } from '@snapshot-labs/sx/dist/utils/encoding';
 import type { Space, SpaceMetadata, Proposal, Vote, User, Choice, NetworkID } from '@/types';
 
-export type PaginationOpts = { limit: number; skip?: number };
+export type PaginationOpts = { limit: number; skip?: number; filter?: any };
```
Sekhmet, line 6: Not really belongs in `PaginationOpts`, for proposal search I just added 3rd optional parameter for search (in that case it's just string for single property instead of dynamic object though).

src/components/NavSettings.vue
```diff
@@ -0,0 +1,45 @@
...
+    }"
+  >
+    <div class="h-[72px] border-b" />
+    <div class="py-4">
+      <router-link
+        v-for="(item, key) in navigationItems"
+        :key="key"
+        :to="{ name: `settings-${key}` }"
+        class="px-4 py-[6px] space-x-2 text-skin-text flex items-center"
+        :class="route.name === `settings-${key}` && 'text-skin-link'"
```
Sekhmet, line 38: Would be object notation for `:class` be cleaner there?

src/components/SpaceItem.vue
```diff
@@ -0,0 +1,33 @@
+<script lang="ts" setup>
+import type { Space as SpaceType } from '@/types';
```
Sekhmet, line 2: I started dropping `import type` convention for pure type imports recently, because it started feeling cumbersome, with not much of benefit. Unless you have preference for those kinds of imports, we should start dropping those from new code.

Sekhmet (commented):
- We should show some message if there are no spaces you control (take a look at other no results messages we have in the app).
- I think UI should react to switching the account in Metamask.

Sekhmet:
> we got overlapping logic in spacesStore and useSpaces composable now. spacesStore is made to use only one spaces list so I moved logic to composable, so maybe we can use the same composable from store?

If it can be integrated nicely, sure.

src/App.vue
```diff
@@ -20,6 +20,12 @@ provide('web3', web3);
 
 const skin = computed(() => userSkin.value);
 const scrollDisabled = computed(() => modalOpen.value || uiStore.sidebarOpen);
+const backdropLeft = computed(
+  () =>
+    `${
+      72 + (['space', 'settings'].includes(route.matched[0]?.name) && !route.meta.hideNav ? 240 : 0)
+    }px`
+);
```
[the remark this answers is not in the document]
Sekhmet: image

Even GitHub's syntax highlighter can't cope with it :D

src/components/NavSpace.vue
```diff
@@ -0,0 +1,68 @@
...
+import IHGlobeAlt from '~icons/heroicons-outline/cash';
+import IHNewspaper from '~icons/heroicons-outline/newspaper';
+import IHCash from '~icons/heroicons-outline/cash';
+import IHCog from '~icons/heroicons-outline/cog';
+
+const route = useRoute();
+const uiStore = useUiStore();
+const spacesStore = useSpacesStore();
+
+// TODO if we not get ALL spaces like paginate 1000 we will not find a space here
```
Sekhmet, line 17: I'm not sure I understand, it's not supposed to get the space from listing all spaces, but when it's fetched at Space view:
https://github.com/snapshot-labs/sx-ui/blob/07233f60a2fa9138ff362633a0a560ab9af7ed3c/src/views/Space.vue#L14-L16

Sekhmet (commented):
If I close sidebar on Settings page hamburger icon (and title) is not visible:
image

src/components/NavSpace.vue
j0hnfl0w: We want to know if space got a treasury in NavSpace so thats why we try to find it in all spaces

j0hnfl0w: ok and it will find it cos fetchSpace() adds space to networksMap.spaces

src/views/Settings.vue
```diff
@@ -1,19 +1,29 @@
...
   <div>
     <div>
       <div
-        class="ml-0 lg:ml-[240px] mr-0 xl:mr-[240px]"
-        :class="{ 'translate-x-[240px] lg:translate-x-0': uiStore.sidebarOpen }"
+        class="ml-0 lg:translate-x-0"
+        :class="{
+          'translate-x-[240px]': !route.meta.hideNav && uiStore.sidebarOpen,
+          'lg:ml-[240px] xl:mr-[240px]': !route.meta.hideNav
+        }"
```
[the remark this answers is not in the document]
j0hnfl0w: there are a lot of patches with paddings and margin for navigations, ideally it would be to use layouts for this and not single app.vue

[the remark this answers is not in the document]
Sekhmet: This particular styling is used exactly like this in two places currently. I'd say it makes sense to put it in single place (as component, or maybe just a helper function that takes hideNav/sidebarOpen.

src/views/Create.vue
```diff
@@ -198,7 +198,7 @@ watch(selectedNetworkId, () => {
           ref="pagesRefs"
           :key="page.id"
           :disabled="!accessiblePages[page.id]"
-          class="px-3 py-1 block lg:w-full rounded text-left scroll-mr-3 first:ml-auto last:mr-auto"
+          class="px-3 py-1 block lg:w-full rounded text-left scroll-mr-3 first:ml-auto last:mr-auto whitespace-nowrap"
```
j0hnfl0w, line 201: Added whitespace-nowrap because of this
image

src/views/Settings.vue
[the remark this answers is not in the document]
j0hnfl0w: Found unused Layout component and used it to wrap pages

Sekhmet (changes requested):
Let's get rid of any sidebar related logic from this PR, as this is not a way to solve it, and just getting it on master moves us into wrong direction.

Problematic commit that introduced those issues will be reverted (https://github.com/snapshot-labs/sx-ui/pull/560)

j0hnfl0w:
Opened new PR for My spaces page
https://github.com/snapshot-labs/sx-ui/pull/561
Maybe better close this PR

PR #509, opened 2023-03-22
feat: add v0.4.0 contracts support

Sekhmet:
## Summary

This PR updates code to work with latest contracts.

Missing functionality (will be added in next PRs, too many changes in this one already):
- No proper proposal validation support (you can only create spaces with vanilla validation, editor checks against proposal threshold that is hardcoded to 0 at API).
- Users can't pick execution strategy if there are multiple
- ~There is no vanilla execution strategy right now (needs to be deployed and some changes are needed on sx-subgraph).~ won't be used

## Test plan

I tested EthSig, Vanilla voting, CompVoting, AvatarExecution.

- Create your own space (you need to use avatar execution, you will only need to enableModule through Gnosis to enable your execution strategy, no other configuration needed).
- Propose
- Vote
- Execute

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "@snapshot-labs/pineapple": "^0.1.0-beta.1",
     "@snapshot-labs/prettier-config": "^0.1.0-beta.7",
     "@snapshot-labs/snapshot.js": "^0.4.11",
-    "@snapshot-labs/sx": "^0.1.0-beta.25",
+    "@snapshot-labs/sx": "^0.1.0-beta.26",
     "@vueuse/core": "^9.4.0",
     "@walletconnect/client": "^1.7.8",
     "ajv": "^8.11.0",
--- a/src/components/Block/SpaceFormStrategies.vue
+++ b/src/components/Block/SpaceFormStrategies.vue
@@ -2,12 +2,18 @@
 import { ref, computed, Ref } from 'vue';
 import type { StrategyConfig, StrategyTemplate } from '@/networks/types';
 
-const props = defineProps<{
-  modelValue: StrategyConfig[];
-  title: string;
-  description: string;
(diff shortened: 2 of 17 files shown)


bonustrack:
This condition is failing for me cuz the vars are string, I believe we should compare them using BigDecimal
https://github.com/snapshot-labs/sx-ui/blob/b6a9c1edf58947c6eb244a666248fd44ee52fb9b/src/views/Proposal.vue#L158
Also I think this condition miss voting period

bonustrack:
Calculating voting power take 3-4 sec for me, any idea what explain this?

bonustrack (approved):
tACK

bonustrack:
Query at specific block with archive node is usually not really slower, it's fast for me here:
https://snapshot.org/#/playground/comp-like-votes?query=eyJwYXJhbXMiOnsic3ltYm9sIjoiQkFMIiwiYWRkcmVzcyI6IjB4MzU4N2IyZjdlMGUyZDYxNjZkNmMxNDIzMGU3ZmUxNjAyNTJiMGJhNCIsImRlY2ltYWxzIjoxOH0sIm5ldHdvcmsiOiI1Iiwic25hcHNob3QiOiI4NzA1MjEwIiwiYWRkcmVzc2VzIjpbIjB4ZUY4MzA1RTE0MGFjNTIwMjI1REFmMDUwZTJmNzFkNWZCY0M1NDNlNyJdfQ
Might come from Infura performance or from the look up of snapshot block

Sekhmet:
Both requests take ~1.5 seconds:
image

But we have to do them serially as they depend on each other.

bonustrack:
@Sekhmet Maybe timestamp to block number is something SX-API could do instead when indexing, but not sure it's something standardized across strategies

PR #511, opened 2023-03-23
chore: use StaticJsonRpcProvider

bonustrack:
Using StaticJsonRpcProvider avoid querying `eth_chainId` before each `eth_call`

Here is an example on proposals page:

Before:
image

After:
image


Sekhmet (approved):
tACK

PR #512, opened 2023-03-23
feat: add website icon

bonustrack:
Add website icon on space overview:
image
http://localhost:8080/#/gor:0x7363dbc13ee217b7725559d53e7b3b089b82ea31

And reduce space between nav items.


bonustrack:
@Sekhmet It should be white on mouse hover ideally but we will need to use SVG directly otherwise we wont be able to fill color

PR #513, opened 2023-03-23
feat: add Proposal validation to create space form

Sekhmet:
## Summary

This PR adds simple UI for configuring Proposal validation for space.

After picking proposal validation you will need to configure it (in case of Voting Power Proposal Validation), by configuring voting power strategies that will be used for it.

**This PR doesn't validate proposal threshold correctly on editor page yet, will be added once sx-subgraph indexes those.**

## Test plan

- Create space with proposal validation
- Propose **This will only work if your voting strategies for voting are the same for proposal validation, if not check below how to hard code it.**

Override strategies here to match your (indexes has to be correct, same as when deployed):
https://github.com/snapshot-labs/sx-ui/blob/752d0027739e2c64d81c2c5dd2d79162aba6818a/src/networks/evm/actions.ts#L128
```js
strategies: [
  {
    index: 0,
    address: '0xeba53160c146cbf77a150e9a218d4c2de5db6b51'
  },
  {
    index: 1,
    address: '0x343baf4b44f7f79b14301cfa8068e3f8be7470de'
  }
],
 ```

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "@braintree/sanitize-url": "^6.0.0",
     "@ensdomains/eth-ens-namehash": "^2.0.15",
     "@esbuild-plugins/node-globals-polyfill": "^0.1.1",
+    "@ethersproject/abi": "^5.7.0",
     "@ethersproject/address": "^5.6.1",
     "@ethersproject/bignumber": "^5.6.2",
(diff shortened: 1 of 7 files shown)


bonustrack (approved):
tACK

PR #514, opened 2023-03-24
feat: use real proposal validation

Sekhmet:
## Summary

This PR adds real proposal validation at propose time (uses correct proposal validation strategies, and computes threshold requirement correctly). It also displays proposal validation in Settings.

**This PR disabled Starknet network, as we would need sx-api changes to make it work, we want to disable SN for now anyway, so keeping it this way for now.**

## Test plan

- Create space with proposal validation (you can use [this PR](https://github.com/snapshot-labs/sx-ui/pull/513))
- Propose works.


bonustrack (approved):
tACK

PR #516, opened 2023-03-24
feat: add vote buttons titles

Sekhmet:
## Summary

This adds titles to buttons that adds more context what they mean.

## Test plan

- Hover on vote buttons.
- You get tooltip.


bonustrack (approved):
tACK

PR #517, opened 2023-03-24
fix: limit proposal link width

Sekhmet:
## Summary

Previously it was possible to click on Proposal's component background to be navigated to proposal (only top half worked), and it was super confusing.

This PR limits clickable area to only proposal title.


bonustrack (approved):
tACK

Issue #506, opened 2023-03-19
style: sticky label on mobile


bonustrack:
The sticky labels on mobile have are one pixel too low, this make content behind appear:
![Screenshot_20230320_004715_Chrome](https://user-images.githubusercontent.com/16245250/226196327-c5e244f4-a64f-4ac2-a795-7caab70e1a1e.jpg)

PR #519, opened 2023-03-24
fix: shift sticky to avoid gap

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/506

Sticky is [bugy](https://twitter.com/youyuxi/status/1436026679059787781?lang=en) and not following screen exactly, which can cause some gaps.

Using workaround to shift it further up to avoid it.

### DIFF
--- a/src/components/Label.vue
+++ b/src/components/Label.vue
@@ -9,7 +9,7 @@ defineProps({
   <h4
     class="eyebrow border-b py-2 px-4 text-skin-text"
     :class="{
-      'sticky top-[72px] bg-skin-bg': sticky
+      'sticky top-[71.5px] bg-skin-bg': sticky
     }"
     v-text="label"
   />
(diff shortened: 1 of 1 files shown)


src/components/Label.vue
```diff
@@ -9,7 +9,7 @@ defineProps({
   <h4
     class="eyebrow border-b py-2 px-4 text-skin-text"
     :class="{
-      'sticky top-[72px] bg-skin-bg': sticky
+      'sticky top-[71.5px] bg-skin-bg': sticky
```
bonustrack, line 12: This seem to have the same effect than using 71px directly, I can see on desktop that the "Voting" label will move slightly on Settings page when scrolling. Can't this be solve by doing 72px on large screen and 71px on small?

Sekhmet: Pushed commit to do that, but you will still have that slight jump on mobile still.

It's a issue in Chrome on HiDPI screens, alternative would be to implement it with JS to switch position automatically, way less elegant but shouldn't have those issues.

bonustrack (approved):
This works better for me, tACK

PR #520, opened 2023-03-24
fix: restore contract modal contents

Sekhmet:
Something changed and now, when you try to edit contract call it won't restore previous values (it will reload ABI).

This PR fixes that, by centralizing `ignoreFormUpdates` in single watcher, that handles both methods that depend on it (handleChangeMethod, handleChangeTo).

## Test plan

- Go to editor.
- Create contract call execution.
- Click edit icon.
- Your execution is restored in the modal.


bonustrack (approved):
tACK

PR #521, opened 2023-03-26
chore: change sx-api URL

bonustrack:
Change SX-API URL from https://testnet-api-1.snapshotx.xyz to https://api-1.snapshotx.xyz to match with UI.


Sekhmet (approved):
utACK

PR #525, opened 2023-03-27
fix: hide nav in nested space routes

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/492

### DIFF
--- a/src/components/Nav.vue
+++ b/src/components/Nav.vue
@@ -23,6 +23,11 @@ const space = computed(() =>
 );
 const { treasury } = useTreasury(space);
 
+const showNavigationItems = computed(() => {
+  if (currentRouteName.value === 'space') return Object.keys(route.params).length === 1;
+  return true;
+});
+
 const navigationConfig = computed(() => ({
   space: {
     overview: {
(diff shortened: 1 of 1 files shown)


src/components/Nav.vue
```diff
@@ -23,6 +23,11 @@ const space = computed(() =>
 );
 const { treasury } = useTreasury(space);
 
+const showNavigationItems = computed(() => {
```
bonustrack, line 26: We should not see the menu area in left and right, the proposal should appear centered on the page and top nav area with Snapshot X should take the whole width (minus sidebar) similar than before. This looks like a trick to hide the menu we should have a more elegant way to do this where we don't run the code of the menu.

j0hnfl0w: agreed we need more elegant solution, need to break Nav to NavSettings and NavSpace and control them better. This PR was like a quick patch, but now it will evolve.

bonustrack: We should avoid quick patch, we only use quick patch in production to fix something urgent

PR #527, opened 2023-03-28
fix: default to null for wallet in edit space modal

Sekhmet:
## Summary

Currently when opening edit space modal for space without treasury configured it will show wallet address field, even though it shouldn't do that.

That's because we are defaulting to '' and when decoding we might put undefined values into them as well.

This PR changes those fields to default to null.

## Test plan

- Go to space you control without treasury (create one or edit existing one if you don't have one).
- Click on edit space button.
- Treasury address field is invisible when No treasury is selected.


j0hnfl0w (approved):
tACK

Issue #539, opened 2023-03-28
feat: align votes labels in Votes modal


Sekhmet:
Currently they are right aligned:
![image](https://user-images.githubusercontent.com/1968722/228266339-34abda3d-c6db-410c-af0b-5a2e59ad5663.png)

Would be nicer if we aligned them to center, within small container on the right side.

PR #541, opened 2023-03-29
feat: align Vote type label in modal to center

Sekhmet:
Closes https://github.com/snapshot-labs/sx-ui/issues/539

### Before

image

## Test plan

- Open votes modal on http://localhost:8080/#/gor:0x928d54f69d15b7dfdac297ca2e1785325e29d653/proposal/1

### DIFF
--- a/src/components/Modal/Votes.vue
+++ b/src/components/Modal/Votes.vue
@@ -39,23 +39,25 @@ watch(open, async () => {
         <div
           v-for="(vote, i) in votes"
           :key="i"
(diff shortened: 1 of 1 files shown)


bonustrack:
I think we should not change the align here. Does this come from SX testing review?

Sekhmet:
@bonustrack From John's review on previous PR, some results look off currently, like:
image

bonustrack (changes requested):
I prefer if we keep the way is it currently

Issue #528, opened 2023-03-28
feat: format small values as ~0


Sekhmet:
Small amounts of voting power should be formatted as `~0`. Similarly balances should be formatted this way too.

PR #543, opened 2023-03-29
feat: add _c helper for formatting currencies

Sekhmet:
This PR adds `_c` helper for currency formatting, that specifies maximum number of decimal digits and displays `~0` in case number would is too small.

Closes: https://github.com/snapshot-labs/sx-ui/issues/528

## Test plan

- Set your VP to small amount (`1n`) here: https://github.com/snapshot-labs/sx-ui/blob/2753c894aa2065e50f9430833f0ac3a743d065b9/src/components/VotingPowerIndicator.vue#L37
- It displays as ~0.

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -68,8 +68,9 @@ export function _n(value: any, notation: 'standard' | 'compact' = 'standard') {
 }
 
 export function _c(value: string | bigint, decimals = 18) {
-  const parsed = Number(value) / 10 ** decimals;
-  if (parsed < 0.000001) return `~0`;
+  const raw = BigInt(value);
+  const parsed = Number(raw) / 10 ** decimals;
+  if (raw !== 0n && parsed < 0.000001) return `~0`;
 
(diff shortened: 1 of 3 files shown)


src/helpers/utils.ts
```diff
@@ -67,6 +67,14 @@ export function _n(value: any, notation: 'standard' | 'compact' = 'standard') {
   return formatter.format(value);
 }
 
+export function _c(value: string | bigint, decimals = 18) {
+  const parsed = Number(value) / 10 ** decimals;
+  if (parsed < 0.000001) return `~0`;
+
+  const formatter = new Intl.NumberFormat('en', { maximumFractionDigits: 6 });
```
bonustrack, line 74: I recommend use 3 digits here, 6 decimals it's quite a lot insignificant, even for BTC
```suggestion
  const formatter = new Intl.NumberFormat('en', { maximumFractionDigits: 3 });
```

bonustrack (approved):
utACK

Issue #526, opened 2023-03-27
chore: add unplugin-auto-import


bonustrack:
Similar than on https://github.com/snapshot-labs/snapshot/pull/3650 we should consider adding `unplugin-auto-import`

PR #546, opened 2023-04-03
feat: add auto-import plugin

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/526

### DIFF
--- a/src/components/S/IAddress.vue
+++ b/src/components/S/IAddress.vue
@@ -5,15 +5,15 @@ export default {
 </script>
 
 <script setup lang="ts">
-type Props = {
-  showPicker?: boolean;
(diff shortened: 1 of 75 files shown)


src/components/S/IAddress.vue
```diff
@@ -5,15 +5,15 @@ export default {
 </script>
```
bonustrack, line 5: Any reason we have script doubled here and on others files?

Sekhmet: We need to export inheritAttrs and it's only possible in non-setup script, so in this case we need one regular script and other (typical) setup script.

bonustrack (approved):
tACK

Issue #501, opened 2023-03-14
feat: voting strategy and space should have symbol


bonustrack:
We should have a symbol for each voting strategies but also for the combined voting strategies. These could be stored within the voting strategy metadataURI and space metadata URI.

PR #547, opened 2023-04-03
feat: add voting power symbols

Sekhmet:
This PR adds symbols for voting power strategies and global voting power symbol (for space).

Closes: https://github.com/snapshot-labs/sx-ui/issues/501

## Test plan

- Create space with custom symbol and with voting strategies that have their own symbol too.
- Go to Proposals page, you can see proper units used.

### DIFF
--- a/src/components/Block/SpaceFormProfile.vue
+++ b/src/components/Block/SpaceFormProfile.vue
@@ -66,8 +66,8 @@ const definition = computed(() => {
       votingPowerSymbol: {
         type: 'string',
         maxLength: 6,
-        title: 'Voting Power symbol',
-        examples: ['VP']
+        title: 'Voting power symbol',
+        examples: ['e.g. VP']
       },
       walletNetwork: {
         type: ['string', 'null'],
--- a/src/components/Block/SpaceFormProfile.vue
+++ b/src/components/Block/SpaceFormProfile.vue
@@ -63,6 +63,12 @@ const definition = computed(() => {
         title: 'Discord',
         examples: ['Discord handle']
       },
+      votingPowerSymbol: {
+        type: 'string',
+        maxLength: 6,
+        title: 'Voting Power symbol',
+        examples: ['VP']
+      },
       walletNetwork: {
         type: ['string', 'null'],
         enum: [null, ...enabledNetworks],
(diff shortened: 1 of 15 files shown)


src/components/Block/SpaceFormProfile.vue
```diff
@@ -63,6 +63,12 @@ const definition = computed(() => {
         title: 'Discord',
         examples: ['Discord handle']
       },
+      votingPowerSymbol: {
+        type: 'string',
+        maxLength: 6,
+        title: 'Voting Power symbol',
```
bonustrack, line 69: We should always use sentence case
```suggestion
        title: 'Voting power symbol',
```

src/views/Create.vue
```diff
@@ -72,6 +72,7 @@ const metadataForm: SpaceMetadata = reactive(
     twitter: '',
     github: '',
     discord: '',
+    votingPowerSymbol: 'VP',
```
bonustrack, line 75: We should leave this field empty by default

src/networks/starknet/actions.ts
```diff
@@ -227,7 +227,7 @@ export function createActions(
       return Promise.all(
         strategiesAddresses.map(async (address, i) => {
           const strategy = getStarknetStrategy(address, defaultNetwork);
-          if (!strategy) return { address, value: 0n, decimals: 0 };
+          if (!strategy) return { address, value: 0n, decimals: 0, token: null, symbol: 'VP' };
```
bonustrack, line 230: The symbol should be empty by default

src/networks/evm/constants.ts
```diff
@@ -163,6 +183,12 @@ export const EDITOR_VOTING_STRATEGIES = [
           type: 'integer',
           title: 'Decimals',
           examples: ['18']
+        },
+        symbol: {
+          type: 'string',
+          maxLength: 6,
+          title: 'Symbol',
+          examples: ['COMP']
```
bonustrack, line 191: We should use `e.g.` on placeholders examples:
```suggestion
          examples: ['e.g. COMP']
```

bonustrack (commented):
Vanilla strategy should also have option to set a symbol

bonustrack (approved):
tACK

PR #548, opened 2023-04-04
feat: add mana-based execution of proposal

Sekhmet:
## Summary

With this PR executing proposals on EVM will be done via mana.

Depends on: https://github.com/snapshot-labs/mana/pull/94

## Test plan

- Use local version of mana: https://github.com/snapshot-labs/mana/pull/94
- Create proposal with execution
- Execute it, it doesn't ask to send you the transaction, instead if does it via mana.


bonustrack (approved):
utACK

Issue #522, opened 2023-03-26
feat: support markdown for proposal body


bonustrack:
We should add support for markdown on proposal and increase the char limit to 6400, for now we should only support these:

### Title h3

**Bold**

_Italic_

> Quote

[Link](url)

This follows what's possible on Medium currently.

PR #549, opened 2023-04-04
feat: add Markdown support to proposal body

Sekhmet:
## Summary

This PR adds markdown support to proposal body. You can now display preview of your proposal in editor as well.

Once submitted proposal will be displayed with Markdown parsed.

Closes: https://github.com/snapshot-labs/sx-ui/issues/522

## Test plan

- Create proposal with Markdown
- View this proposal.


src/components/Markdown.vue
```diff
@@ -0,0 +1,202 @@
...
+
+const parsed = computed(() => remarkable.render(props.body));
+</script>
+
+<template>
+  <div class="markdown-body break-words" v-html="parsed" />
+</template>
+
+<style lang="scss">
+.markdown-body {
```
Sekhmet, line 24: Styles from Snapshot.

j0hnfl0w (approved):
tACK

bonustrack:
Can you force to have only the styles specified in the issue?

src/views/Editor.vue
```diff
@@ -159,7 +160,22 @@ watch(proposalData, () => {
           title: 'Title'
         }"
       />
-      <div class="s-base mb-5">
+      <div class="flex">
+        <Link
+          :is-active="!previewEnabled"
+          text="Editor"
```
bonustrack, line 166: How about using same naming than GitHub? Editor is the whole page not just the body
```suggestion
          text="Write"
```

bonustrack:
@Sekhmet Correct

Sekhmet:
@bonustrack doesn't seem possible with remarkable, we can enable global features like headings, but not specific ones:
I could enable:
- headings (# - ######)
- links (both links and images)
- emphasis (bold and itallic)
- blockquote

bonustrack:
@Sekhmet Ok sound good, maybe just make it so h1 and h2 use the same font-size in the proposal body. I noticed that `remarkable` seem to not be that popular anymore, was not updated for quite some time we may want to replace it with another one at some point.

PR #554, opened 2023-04-11
feat: add avatar upload for space edit

j0hnfl0w:
Closes: https://github.com/snapshot-labs/sx-ui/issues/419
Moved from fork PR: https://github.com/snapshot-labs/sx-ui/pull/429

- [x] uploads avatar to ipfs with pineapple
- [x] save avatar url to space.avatar, need to be an owner of the space => need to create my space
- [x] add sx resolver to stamp.fyi service to serve sx space avatars
       like a space resolver [here](https://github.com/snapshot-labs/stamp/blob/master/src/resolvers/space.ts)
- [ ] upload avatar on space creation

### DIFF
--- a/src/components/Modal/EditSpace.vue
+++ b/src/components/Modal/EditSpace.vue
@@ -1,9 +1,11 @@
 <script setup lang="ts">
 import { clone } from '@/helpers/utils';
 import type { Space, SpaceMetadata, NetworkID } from '@/types';
+import SIUploadImage from '@/components/S/IUploadImage.vue';
 
 const DEFAULT_FORM_STATE: SpaceMetadata = {
   name: '',
+  avatar: '',
   description: '',
   externalUrl: '',
   twitter: '',
@@ -28,6 +30,7 @@ const searchValue = ref('');
 const sending = ref(false);
 const formErrors = ref({} as Record<string, string>);
 const form: SpaceMetadata = reactive(clone(DEFAULT_FORM_STATE));
+const avatarInput = ref<InstanceType<typeof SIUploadImage> | null>(null);
 
 function handlePickerSelect(value: string) {
   showPicker.value = false;
@@ -38,6 +41,13 @@ async function handleSubmit() {
   sending.value = true;
 
   try {
+    // Upload avatar
+    console.log('avatarInput', avatarInput.value);
+    console.log('avatarInput?.value?.previewFile?', avatarInput?.value?.previewFile);
+    if (avatarInput?.value?.previewFile) {
+      await avatarInput?.value?.uploadFile();
+    }
+
     await updateMetadata(props.space, form);
     emit('close');
   } finally {
@@ -51,6 +61,7 @@ watch(
     showPicker.value = false;
 
     form.name = props.space.name;
+    form.avatar = props.space.avatar;
     form.description = props.space.about || '';
     form.externalUrl = props.space.external_url;
     form.github = props.space.github;
@@ -91,6 +102,31 @@ watch(
         </div>
       </template>
     </template>
+    <div v-if="!showPicker" class="relative bg-skin-border h-[100px] -mb-[50px]" />
+    <SIUploadImage
+      v-if="!showPicker"
+      ref="avatarInput"
+      class="relative h-[80px] ml-4 group max-w-max cursor-pointer"
+      @image-uploaded="url => (form.avatar = url)"
+    >
+      <template #avatar="{ uploading, previewUrl }">
+        <Stamp
+          :id="space.id"
+          :size="80"
+          class="pointer-events-none border-[4px] border-skin-bg !bg-skin-bg !rounded-lg group-hover:opacity-80"
+          :class="{
+            'opacity-80': uploading
+          }"
+        />
+        <div
+          class="pointer-events-none absolute group-hover:visible inset-0 z-10 flex flex-row w-full h-full items-center content-center justify-center"
+        >
+          <IH-pencil v-if="!uploading" class="invisible text-skin-link group-hover:visible" />
+          <UiLoading v-if="uploading" class="block bg-green z-5" />
+          <img v-if="previewUrl" :src="previewUrl" class="absolute w-full h-full z-3 !rounded-lg" />
+        </div>
+      </template>
+    </SIUploadImage>
     <BlockContactPicker
       v-if="showPicker"
       :loading="false"
--- a/src/components/S/IUploadImage.vue
+++ b/src/components/S/IUploadImage.vue
@@ -1,0 +1,63 @@
+<script setup lang="ts">
+const emit = defineEmits(['image-uploaded', 'image-remove']);
+
+const { upload, isUploadingImage } = useImageUpload();
+
+const fileInput = ref<HTMLInputElement | null>(null);
+function openFilePicker() {
+  fileInput.value?.click();
+}
+
+const uploadSuccess = ref(false);
+const previewFile = ref<File | undefined>(undefined);
+const previewUrl = ref<string | undefined>(undefined);
+
+function handleFileChange(e: Event) {
+  uploadSuccess.value = false;
+  if ((e.target as HTMLInputElement).files?.[0]) {
+    previewFile.value = (e.target as HTMLInputElement).files?.[0];
+    previewUrl.value = URL.createObjectURL(previewFile.value as File);
+  }
+}
+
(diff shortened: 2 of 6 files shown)


j0hnfl0w:
Added PR for stamp: https://github.com/snapshot-labs/stamp/pull/51

j0hnfl0w:
@Sekhmet how can we add avatar to Space type [here in subgraph](https://api.thegraph.com/subgraphs/name/snapshot-labs/sx-goerli)?

src/composables/useImageUpload.ts
```diff
@@ -0,0 +1,58 @@
+import { ref } from 'vue';
+import { upload as pin } from '@snapshot-labs/pineapple';
+
+const isUploadingImage = ref(false);
```
Sekhmet, line 4: Do we want to make it global?

src/components/Modal/EditSpace.vue
```diff
@@ -91,6 +102,31 @@ watch(
...
       </template>
     </template>
+    <div v-if="!showPicker" class="relative bg-skin-border h-[100px] -mb-[50px]" />
+    <SIUploadImage
+      v-if="!showPicker"
+      ref="avatarInput"
+      class="relative h-[80px] ml-4 group max-w-max cursor-pointer"
+      @image-uploaded="url => (form.avatar = url)"
+    >
+      <template #avatar="{ uploading, previewUrl }">
```
Sekhmet, line 112: uploading prop doesn't seem to do anything, I see no difference when image is being uploaded.

Sekhmet (commented):
It seems overly complex, for what we need in my opinion. Can we simplify it?

1. I feel like adding extra composable makes it more complex than it needs to be now.
2. If we want to keep delayed image upload I think we can just make it so our component just exposes File and we can let consumer upload it themselves, keeping it much simpler.
3. If we want upload image as it's updated, I think that's okay too, and removes complexity.

src/components/S/IUploadImage.vue
```diff
@@ -0,0 +1,63 @@
+<script setup lang="ts">
+const emit = defineEmits(['image-uploaded', 'image-remove']);
+
+const { upload, isUploadingImage } = useImageUpload();
+
+const fileInput = ref<HTMLInputElement | null>(null);
+function openFilePicker() {
+  fileInput.value?.click();
+}
```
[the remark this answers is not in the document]
j0hnfl0w: I understand this is our codestyle here, but we are still organizing code just like it is Options API
https://vuejs.org/guide/extras/composition-api-faq.html#more-flexible-code-organization

j0hnfl0w:
> @Sekhmet how can we add avatar to Space type [here in subgraph](https://api.thegraph.com/subgraphs/name/snapshot-labs/sx-goerli)?

@Sekhmet can you help with this?

Sekhmet:
> @Sekhmet can you help with this?

https://github.com/snapshot-labs/sx-subgraph/pull/16

j0hnfl0w:
Now need to review and merge https://github.com/snapshot-labs/stamp/pull/51 for avatars to work.

Sekhmet:
@j0hnfl0w can you remove draft status once it's ready to review?

src/components/S/IStamp.vue
```diff
@@ -0,0 +1,80 @@
...
+
+function handleFileChange(e: Event) {
+  uploadSuccess.value = false;
+  const file = (e.target as HTMLInputElement).files?.[0];
+  if (file) {
+    upload(file, image => {
+      uploadSuccess.value = true;
+      emit('update:modelValue', image.url);
+    });
+  }
```
Sekhmet, line 38:
```suggestion
  if (!file) return
  
  upload(file, image => {
    uploadSuccess.value = true;
    emit('update:modelValue', image.url);
  });
  }
```

src/composables/useImageUpload.ts
```diff
@@ -0,0 +1,55 @@
...
+  const imageName = ref('');
+
+  const reset = () => {
+    isUploadingImage.value = false;
+    imageUploadError.value = '';
+    imageUrl.value = '';
+    imageName.value = '';
+  };
+
+  const upload = async (
```
Sekhmet, line 17: What do you think of making it purely async, instead of having callback as well?

src/components/S/IStamp.vue
```diff
@@ -0,0 +1,80 @@
...
+  modelValue?: string;
+  error?: string;
+  definition: any;
+}>();
+
+const emit = defineEmits<{
+  (e: 'update:modelValue', value: string);
+}>();
+
+const { upload, isUploadingImage } = useImageUpload();
```
Sekhmet, line 14: Seems a bit weird, that isUploadingImage is part of the composable, but uploadSuccess isn't.

src/composables/useImageUpload.ts
```diff
@@ -0,0 +1,55 @@
...
+    isUploadingImage.value = true;
+    const formData = new FormData();
+
+    // TODO: Additional Validations - File Size, File Type, Empty File, Hidden File
+    if (!['image/jpeg', 'image/jpg', 'image/png'].includes(file.type)) {
+      imageUploadError.value = 'Unsupported file type';
+      isUploadingImage.value = false;
+      return;
+    }
+    formData.append('file', file);
```
Sekhmet, line 32:
```suggestion
    // TODO: Additional Validations - File Size, File Type, Empty File, Hidden File
    if (!['image/jpeg', 'image/jpg', 'image/png'].includes(file.type)) {
      imageUploadError.value = 'Unsupported file type';
      isUploadingImage.value = false;
      return;
    }
    
    const formData = new FormData();
    formData.append('file', file);
```

`formData`'s initialization and set up looks weird being separated by unrelated function.

src/views/Space/Overview.vue
```diff
@@ -78,6 +78,7 @@ const grouped = computed(() => {
           <Stamp
             :id="space.id"
             :size="90"
+            type="space-sx"
```
Sekhmet, line 81: Is this supposed to work with this PR?
https://github.com/snapshot-labs/stamp/pull/51/files

I don't see how this type ends up being used.

https://github.com/snapshot-labs/stamp/blob/f049ba2bd614729ece6b363195477bd8e5ddbddc/src/api.ts#L55-L57

j0hnfl0w: Yeah took space-sx  by mistake from [here](https://github.com/snapshot-labs/stamp/pull/51/files#diff-d43475a59a1dcd7ce7ecfab266b52211539c3d0148385f07485dd3de61a34964R8)

Sekhmet (commented):
Right now, on space with existing avatar if I upload different image I see loading indicator for a while, but once it disappears I don't see new avatar, only the old one.


https://user-images.githubusercontent.com/1968722/233089282-6b888457-c2fd-4231-8f5a-8d8b659ca65c.mp4

j0hnfl0w:
Can't reproduce this bug, are there errors in network? Maybe the file size is too large.
Need to add notify about errors.

src/composables/useImageUpload.ts
```diff
@@ -0,0 +1,53 @@
+import { ref } from 'vue';
```
Sekhmet, line 1: The way it's currently used I'd say it doesn't make sense to be composable (consumer gets imageUrl/imageName) from the function call, so refs are there for nothing.

It could just be a function that we call to upload image instead, no vue integration needed.

Sekhmet: That part makes some sense, but given that we end up using it just as a regular function (+ reactive isUploadingImage) in the makes me think it's over complicated.

We only need to use that logic in single component currently and it will be much faster as non-reactive helper function where isUploadingImage is handled by customer.

When we end up lacking some functionality with that approach we can go back to it and see if reactivity makes sense.

src/components/S/IStamp.vue
```diff
@@ -0,0 +1,84 @@
...
+<template>
+  <div
+    v-bind="$attrs"
+    class="relative h-[80px] group max-w-max cursor-pointer mb-3"
+    @click="openFilePicker()"
+  >
+    <img
+      v-if="imgUrl"
+      :src="imgUrl"
+      class="w-full min-w-[80px] h-full !rounded-lg"
```
Sekhmet, line 55: Are we missing group-hover here?

Sekhmet (approved):
tACK

PR #552, opened 2023-04-06
feat: add confirmation screen when deploying space

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/452

This PR adds extra screen in editor that shows progress of deployment of the space.

Changes:
- Now deploying dependencies (currently execution strategies) has visual feedback.
- UI waits for previous transaction to confirm, in case they fail.
- You can access transaction link for each step.
- Actual controller from form is used for new space, currently it was hardcoded to person singing the transaction.

PR #556, opened 2023-04-12
feat: add selection of execution strategy at proposal creation time

Sekhmet:
## Summary

Depends on https://github.com/snapshot-labs/sx-ui/pull/552

Added field to make it possible to pick from different voting strategies that are configured for a space.

## Test plan
- Create space with multiple execution strategies, or use this one.
- You can select execution strategy when creating a proposal.

## Screenshots

image

### DIFF
--- a/src/components/Block/ExecutionEditable.vue
+++ b/src/components/Block/ExecutionEditable.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import draggable from 'vuedraggable';
-import { Transaction as TransactionType, Space, SelectedStrategy } from '@/types';
 import { getNetwork } from '@/networks';
+import { Transaction as TransactionType, Space, SelectedStrategy } from '@/types';
 
 const props = defineProps<{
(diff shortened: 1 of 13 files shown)


bonustrack:
I have this error with a blank page when I go to the editor

image

http://localhost:8080/#/gor:0x17ecebb4180529c0e47dada8b6a6d2b9ee92ff4a/create

bonustrack:
@Sekhmet Can you use a spacing of `mb-3` for the block with execution list and use a white border when selected instead of green?

image

Also we should use the default link color for the text and icon when execution is not selected, the current gray color feel like the button is disabled (not clickable).

bonustrack:
One issue i found is that now when we trigger a new proposal from the treasury it wont show the execution directly, this might be confusing but not sure what the best flow here. Perhaps if there is only one execution we can default to it.

bonustrack:
Looks good, can you just increase spacing here with `mb-3`

image

Issue #518, opened 2023-03-24
style: create tooltip component


bonustrack:
We should have our own tooltip components, style can look like this:
<img width="560" alt="image" src="https://user-images.githubusercontent.com/16245250/227538628-606dcf78-69fc-46d2-8ec3-ffa043d6db61.png">

PR #570, opened 2023-04-20
feat: add UiTooltip

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/518

This PR adds UiTooltip component and uses it as replacement for native `title`. **Currently results still use tooltip as they use overflow-hidden to achieve rounded corners, which makes tooltip hidden.**

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -7,7 +7,7 @@
   --bg-color: #1c1b20;
   --text-color: #a09fa5;
   --link-color: #faf9fc;
-  --tooltip-color: #a09fa4;
+  --tooltip-text-color: #a09fa4;
   --content-color: rgba(250, 249, 252, 0.8);
   --heading-color: #faf9fc;
   --border-color: #333237;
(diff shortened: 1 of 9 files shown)


src/style.scss
```diff
@@ -7,6 +7,7 @@
   --bg-color: #1c1b20;
   --text-color: #a09fa5;
   --link-color: #faf9fc;
+  --tooltip-color: #a09fa4;
```
bonustrack, line 10: We don't need to introduce new colors, the color of the tooltip background is exactly the color of borders

src/views/Proposal.vue
```diff
@@ -196,33 +196,36 @@ watch([() => web3.value.account, proposal], () => getVotingPower());
...
+                <UiButton
+                  class="w-full !text-white !bg-red !border-red"
+                  :primary="true"
+                  :loading="sendingType === 2"
+                  @click="handleVoteClick(2)"
+                >
+                  <IH-x class="inline-block" />
+                </UiButton>
+              </UiTooltip>
+              <UiTooltip title="Abstain Abstain Abstain Abstain Abstain">
```
bonustrack, line 219:
```suggestion
              <UiTooltip title="Abstain">
```

bonustrack (commented):
Can you make tooltip to appear on top of the content by default?

bonustrack:
Thanks, I think actually it would be better to use link color for text within the tooltip and give up on new color var. We should also increase space between tooltip and element by 2px, currently it's bit too close together:

Before
image
After
image

bonustrack (approved):
tACK

Issue #380, opened 2023-01-19
feat: add form to edit space settings


bonustrack:
TBD

PR #571, opened 2023-04-20
feat: add settings edit option

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/380
Depends on https://github.com/snapshot-labs/sx.js/pull/189

Allows to edit voting delay, min/max voting period and controller.

UI is likely temporary, as editing those values is only possible one by one currently, so single form doesn't really make sense.

But it works for now.

## Test plan

- Link https://github.com/snapshot-labs/sx.js/pull/189
- Go to settings of space you control.
- Change some settings.


bonustrack:
Looks good, can you just make it so the edit Icon only appear on mouse over?

bonustrack (approved):
tACK

Issue #557, opened 2023-04-18
style: update proposal layout


bonustrack:
We should change the layout of the proposal page to this:
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/232839631-30c5ce3e-3ebd-419b-b78a-9d14ff554872.png">

Some of the changes:
- Title of the proposal should use a font-size of 36px
- The content of the proposal should use link color with 80% transparency
- The width of the content should be 630px

Here is the Figma file:
https://www.figma.com/proto/8DA7eveuzbdgbktIMf3EXu/SX-UI-2?node-id=219-53&scaling=min-zoom&page-id=0%3A1&hide-ui=1

Sekhmet:
@bonustrack questions related to this UI.

1. How it's supposed to look like on mobile?
2. Where do we show execution block?
3. Where do we show actions (execute transactions button).
4. With current design we don't see actual values of VP for each vote https://github.com/snapshot-labs/sx-ui/issues/434
5. What about list of votes. Currently we display `N votes` that you can click to bring it up, do we move it to the right section?
6. What about timeline of the proposal. We currently have `N days left` that you can click to bring it up, do we move it to the right section?
7. With separation of quorum from results bar will be showing different values (currently if quorum is at 10% and 100% votes are FOR results bar will show 10% green, 90% empty, when it's separated results bar will show 100% green). We show the same bar on proposals list, which way of displaying votes should we use there?).

It would be great to see more complex design, with all functionality, ideally with mobile variant.

PR #572, opened 2023-04-21
feat: update Proposal layout

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/557
Depends on: https://github.com/snapshot-labs/sx-ui/pull/570

Currently no separate results/quorum section, pending some questions being answered.

Everything that was not included on the design was kept in main section (actions, execution etc.)

## Screenshot

image

### DIFF
--- a/src/components/Markdown.vue
+++ b/src/components/Markdown.vue
@@ -56,7 +56,7 @@ const parsed = computed(() => remarkable.render(props.body));
 
 <style lang="scss">
 .markdown-body {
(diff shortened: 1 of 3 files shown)


bonustrack:
- The font-size for the content of the proposal should be 22px
- Font-size for title should be 36px
- Right sidebar should have a left border
- "Your voting power", "Cast your vote" and "Results" title should use same style than others labels with `eyebrow` class

Issue #574, opened 2023-04-23
bug: avatar should always be squared


bonustrack:
When uploading an avatar that is not a square it show a rectangle, we should always display a squared image instead.
<img width="506" alt="image" src="https://user-images.githubusercontent.com/16245250/233860545-1eb686ad-5781-4ad9-9485-952a6af0543f.png">

PR #577, opened 2023-04-24
fix: display stamp preview as square

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/574

## Test plan

- Upload rectangle image.
- Preview is squared with cover fit (same as Stamp).


j0hnfl0w (approved):
tACK

Issue #575, opened 2023-04-24
style: add tooltip on every header buttons


bonustrack:
Here are the label for buttons:

Copy, Send token / Send NFT
<img width="1469" alt="image" src="https://user-images.githubusercontent.com/16245250/233945100-86504d81-e9a6-465c-967b-63fff6bb2cc7.png">

New proposal, Add to favorite / Remove from favorite
<img width="1468" alt="image" src="https://user-images.githubusercontent.com/16245250/233946186-b45f962d-792b-4b19-b68d-cb27a43615b0.png">

New proposal
<img width="1465" alt="image" src="https://user-images.githubusercontent.com/16245250/233945832-30a90352-7ac3-4548-8f3e-2c33241559d7.png">

PR #578, opened 2023-04-24
feat: add header buttons tooltips

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/575

@bonustrack currently tooltips blend into Space background because they use the same color. Do you want to use some border on tooltips to prevent it?
image

## Test plan
- Check space header buttons, all should have tooltips.

### DIFF
--- a/src/components/Ui/Tooltip.vue
+++ b/src/components/Ui/Tooltip.vue
@@ -43,7 +43,7 @@ const arrowStyle = computed(() => {
 </script>
 
 <template>
-  <div class="relative">
+  <div class="relative inline-block">
     <div
       ref="referenceRef"
       class="h-full"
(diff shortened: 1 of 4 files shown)


bonustrack:
Space between the 2 buttons should be smaller (same than on others page or on https://snapshotx.xyz)
image

Sekhmet:
> @Sekhmet
> 
> > @bonustrack currently tooltips blend into Space background because they use the same color. Do you want to use some border on tooltips to prevent it?
> 
> Can you try to add a 1px border on the tooltip with the same color than the background to not overlap with cover?

Doesn't look great:
image
image

bonustrack (approved):
tACK, we can solve the issue with cover bg later

PR #583, opened 2023-04-25
fix: keep reactivity for space in useTreasury

Sekhmet:
Current hook usage can lose reactivity, for example refreshing editor page will result in treasury not being found (buttons will be disabled) because space prop is loaded later.
This PR fixes it by making sure we are passing reactive value.

## Test plan
- Go to http://localhost:8080/#/gor:0x65e4329e8c0fba31883b98e2cf3e81d3cdcac780/create
- Refresh page
- Execution buttons are all enabled


bonustrack (approved):
tACK

PR #592, opened 2023-04-26
feat: add strategies descriptions

bonustrack:
Display for strategies descriptions on space creation
image

### DIFF
--- a/src/components/Block/SpaceFormNetwork.vue
+++ b/src/components/Block/SpaceFormNetwork.vue
@@ -18,7 +18,7 @@ const availableNetworks = enabledNetworks.map(id => ({
 
 <template>
   <div class="mb-4">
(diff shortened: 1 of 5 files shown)


Sekhmet (commented):
image

Do you think it would look better centered like this? It looks a bit off with icon on top I think.

@bonustrack CI is failing and it needs a rebase. I can handle it if you want.

bonustrack:
@Sekhmet Centered looks good too feel free to take over this PR. I think also we should enforce a certain height for the card with truncate of text so it doesnt look bad on mobile

Sekhmet (approved):
tACK

Sekhmet:
Added real descriptions for strategies from https://docs.snapshotx.xyz/protocol-sx-evm/authenticators

Issue #591, opened 2023-04-25
bug: allow only unique authenticator strategy


bonustrack:
Currently it's possible to select multiple time a single strategy, this shouldnt be the case as authenticator dont have any parameters.

PR #593, opened 2023-04-26
feat: add unique constraint to authenticator strategies

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/591

## Test plan

- Go to space creation form, authenticators step.
- You can only add one authenticator of each type.


bonustrack (approved):
tACK

PR #595, opened 2023-04-26
feat: display ozVotes token in modal

Sekhmet:
Currently ozVotes strategies do not show connected in VP modal, this can be added with simple fix.

## Test plan
- Go to http://localhost:8080/#/gor:0x90ccccd7667a8f67404c2f6fe3124ced1bb91237/proposals
- Click VP.
- Token address is shown in OZ votes voting strategy.

### DIFF
--- a/src/networks/evm/actions.ts
+++ b/src/networks/evm/actions.ts
@@ -292,7 +292,9 @@ export function createActions(
             provider
           );
 
-          const token = strategy.type === 'comp' ? strategiesParams[i] : undefined;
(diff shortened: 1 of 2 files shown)


bonustrack:
Looks good, can we just use same name than on space creation? It was renamed to `OpenZeppelin Votes` on #592, ideally we should reuse the same text and not duplicate description on multiple place, and ideally all this should endup in SX.js so when a contributor want to create a new strategy he only need to PR SX.js.

bonustrack (approved):
tACK

Issue #610, opened 2023-04-27
bug: tooltip appear at wrong position


bonustrack:
Tooltip appear in wrong position if you load a proposal and mouse over the voting button before your voting power get loaded.

<img width="405" alt="image" src="https://user-images.githubusercontent.com/16245250/234964066-b8f7c698-863e-43db-8431-54c3d90c521c.png">

PR #615, opened 2023-04-28
fix: use relative to set position of Tooltip

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/610

## Test plan
- Tooltips for vote buttons appear at correct position.
- Nothing breaks.


bonustrack (approved):
tACK

PR #616, opened 2023-04-28
fix: move attrs and condition to Tooltip for indicator button

Sekhmet:
## Test plan

- Do some action (vote/create proposal).
- Hover over pending indicator (and keep it until indicator disappears).
- Tooltip's arrow is pointing to the center of the button.
- Once indicator disappears tooltip does too.


bonustrack (approved):
tACK

PR #618, opened 2023-04-28
refactor: preserve types in useActions

Sekhmet:
It's annoying to deal with useActions right now because we lose types, this recovers proper types.


bonustrack (approved):
utACK

Issue #534, opened 2023-03-28
feat: make form validation error messages more meaningful


Sekhmet:
Currently it will only show Invalid field, without more context, we should try to find a solution to generate user friendly messages for those.

PR #624, opened 2023-05-04
feat: add user friendly form errors

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/534

This PR adds user friendly form messages.

1. Tries to use default message if it's not format error message.
2. If it's format error message use custom messages per type.
3. If it's format error message we don't have custom message for we show generic `Invalid format.` message as otherwise format isn't great (`must have format "ethValue"`).

## Test plan

Check some forms trying to break them that you know have validation working.

## Screenshots

image

### DIFF
--- a/src/helpers/validation.ts
+++ b/src/helpers/validation.ts
@@ -13,10 +13,11 @@ function getErrorMessage(errorObject: ErrorObject): string {
       case 'address':
         return 'Must be a valid address.';
       case 'uint256':
+        return 'Must be a positive integer.';
       case 'int256':
-        return `Must be a valid ${errorObject.params.format} value.`;
+        return 'Must be an integer.';
       case 'ethValue':
-        return 'Must be a valid Ethereum value.';
+        return 'Must be a number.';
       default:
         return 'Invalid format.';
     }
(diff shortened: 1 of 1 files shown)


src/helpers/validation.ts
```diff
@@ -1,10 +1,32 @@
...
+
+  if (errorObject.keyword === 'format') {
+    switch (errorObject.params.format) {
+      case 'address':
+        return 'Must be a valid address.';
+      case 'uint256':
+      case 'int256':
+        return `Must be a valid ${errorObject.params.format} value.`;
+      case 'ethValue':
+        return 'Must be a valid Ethereum value.';
```
bonustrack, line 19: This might be confusing, I'd rather say something like:
```suggestion
        return 'Must be a number.';
```

src/helpers/validation.ts
```diff
@@ -1,10 +1,32 @@
...
+function getErrorMessage(errorObject: ErrorObject): string {
+  if (!errorObject.message) return 'Invalid field.';
+
+  if (errorObject.keyword === 'format') {
+    switch (errorObject.params.format) {
+      case 'address':
+        return 'Must be a valid address.';
+      case 'uint256':
+      case 'int256':
+        return `Must be a valid ${errorObject.params.format} value.`;
```
bonustrack, line 17: This also might be more obvious for regular users
```suggestion
      case 'uint256':
        return 'Must be a positive integer.';
      case 'int256':
        return 'Must be an integer.';
```

PR #625, opened 2023-05-04
fix: parse fetched ABI

Sekhmet:
## Summary

Regression from: https://github.com/snapshot-labs/sx-ui/pull/563

## Test plan

- Go to create proposal page.
- Create contract call execution with contract address of `0x3587b2F7E0E2D6166d6C14230e7Fe160252B0ba4`.
- It fetched ABI properly.


bonustrack (approved):
tACK

Issue #606, opened 2023-04-27
feat: add input validation for urls numbers and addresses


bonustrack:
We should validate input for address, urls and number for:

Proposal creation:
- discussion input should be an url
- Send token and NFT address input should be an address

Space creation and edit profile:
- website should be an url
- controller should be an address
- Safe module quorum should be number
- Timelock quorum and delay should be number

PR #626, opened 2023-05-04
feat: add missing form validation

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/606

This PR adds missing form validation:
- URL validation for discussion and website URL
- address validation for send token/NFT recipient and controller
- integer validation for quorum/timelock delay

It also required adding extra `skipEmptyOptionalFields` option to `validateForm` so we can decide what happens if we pass empty value for optional field, in most cases it might make sense to make it valid, which can be achieved with this option.

## Test plan

- Play with website URL, executions arguments and controller in create space form.
- Play with recipients in proposal execution modals and discussion URL.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "@vueuse/core": "^9.4.0",
     "@walletconnect/client": "^1.7.8",
     "ajv": "^8.11.0",
+    "ajv-formats": "^2.1.1",
     "buffer": "^6.0.3",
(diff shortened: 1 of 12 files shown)


bonustrack:
Looks good, can we also validate inputs on settings page?

bonustrack (approved):
tACK

Issue #621, opened 2023-04-29
feat: allows for space or proposal without execution


bonustrack:
It should be possible to create a space without any execution strategy, and it should be possible to create a proposal without any execution (even if the space has execution some strategies)

bonustrack:
To my understanding we need to pass an execution parameter on the contract when we create a proposal, but this param can be empty, right @pscott? Or do we need some smart contract changes?

pscott:
You can definitely create a proposal with an execution strategy set to `0` (so not "empty" but set to 0), but of course you won't be able to call the `execute` function.
We could make a special case where `execute` doesn't actually `execute` anything if `executionStrategy.address == 0` (one `if` condition to add in the space contract).

pscott:
For code context:

1) In the `propose` function we store the `executionStrategy.addr` here: https://github.com/snapshot-labs/sx-evm/blob/b42170b3e50eeefa2c1886536d1bf9fa2c51ae11/src/Space.sol#L373

2) In the `execute` function we call the `execute` method on the contract that we stored previously... if you set it to 0, we will call `execute` on contract located at `0x000` (which doesn't exist ofc): 
https://github.com/snapshot-labs/sx-evm/blob/b42170b3e50eeefa2c1886536d1bf9fa2c51ae11/src/Space.sol#L437

PR #627, opened 2023-05-04
feat: allow proposals without execution

Sekhmet:
## Summary

This PR enables option to create spaces with no execution strategies and proposals without execution.

Closes: https://github.com/snapshot-labs/sx-ui/issues/621

## Test plan

- Deploy space with no execution strategies.
- Create proposal with no execution (in space with no executions).
- Create proposal with no execution (in space with executions).
- Create proposal with execution (in space with executions).
- Edit proposal above to remove that execution (should disappear, quorum will remain as it needs sx-subgraph update).

## Screenshots

image

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -157,7 +157,7 @@ export function useActions() {
     title: string,
     body: string,
     discussion: string,
-    executionStrategy: string,
+    executionStrategy: string | null,
     execution: Transaction[]
   ) {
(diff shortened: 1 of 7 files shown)


bonustrack:
Looks good, can you just remove the block "No execution", we can make that by default there isn't any execution, if user want an execution he can just select one, and if he change his mind we can make it so when he click on an active execution it get unselected.

Issue #609, opened 2023-04-27
style: 1px bug on mobile


bonustrack:
On mobile there is 1px empty between topnav and the header of the page, we should remove that.
![Screenshot_20230427_161142_Chrome](https://user-images.githubusercontent.com/16245250/234950122-eabb6520-0c7d-4cbe-be40-ef453b40689d.jpg)

PR #629, opened 2023-05-04
fix: offset cover photo 1px up to handle overscroll bug

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/609

On mobile scrolling with overscroll might result in inaccurate final position (it's overscrolled at rest by subpixel margin) causing background to be visible where it should be covered entirely by topnav.

This PR only affects Cover because it's the only thing that is currently impacted. If we find it happening in other places we could change app's margin from 72px to 71px to compensate.,

### DIFF
--- a/src/views/Space/Overview.vue
+++ b/src/views/Space/Overview.vue
@@ -56,7 +56,7 @@ const grouped = computed(() => {
 
 <template>
   <div>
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

Issue #597, opened 2023-04-27
feat: input for duration should show days, hours and minutes


bonustrack:
The input with duration like min duration, max duration, timelock delay should be divided in 3 input with days hours and minutes

<img width="640" alt="image" src="https://user-images.githubusercontent.com/16245250/234905511-15c085bf-73cc-4756-a2e2-27ff26abfd6c.png">

j0hnfl0w:
or it can be one date input to select end date with calendar

Sekhmet:
> Can it be one date input to select end date with calendar?

It's duration not a date.

j0hnfl0w:
yeah it's a duration, but from the user perspective can it be like "we want to finish voting on a Sunday evening" so user need to calculate duration to specific date.

bonustrack:
How would that work, this is a global settings for the space not for a specific proposal

j0hnfl0w:
So all proposals in a space has the same duration? If so setting a date is pointless.

j0hnfl0w:
Ah yeah it is for min/max duration, sorry

PR #630, opened 2023-05-04
feat: add SIDuration input

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/597

### DIFF
--- a/src/components/Block/SpaceFormVoting.vue
+++ b/src/components/Block/SpaceFormVoting.vue
@@ -26,18 +26,18 @@ const definition = {
   properties: {
     votingDelay: {
       type: 'number',
-      title: 'Voting delay',
-      examples: ['0']
+      format: 'duration',
+      title: 'Voting delay'
     },
     minVotingDuration: {
       type: 'number',
-      title: 'Min. voting duration',
-      examples: ['0']
+      format: 'duration',
+      title: 'Min. voting duration'
     },
     maxVotingDuration: {
       type: 'number',
-      title: 'Max. voting duration',
-      examples: ['86400']
+      format: 'duration',
+      title: 'Max. voting duration'
     },
     ...(!evmNetworks.includes(props.selectedNetworkId) && {
       proposalThreshold: {
--- a/src/components/S/IDuration.vue
+++ b/src/components/S/IDuration.vue
@@ -34,15 +34,15 @@ watch(
 <template>
   <div>
     <div v-text="definition.title" />
-    <div class="flex gap-3">
+    <div class="flex">
       <SBase :definition="{ title: 'Days' }" class="flex-1">
-        <input v-model="days" class="s-input" type="number" min="0" />
+        <input v-model="days" class="s-input !rounded-r-none" type="number" min="0" />
       </SBase>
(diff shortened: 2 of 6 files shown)


bonustrack:
Can you make it look like this:
image
The height of the grey block should be the same as others inputs (61px) currently it's a bit bigger. Also we should not use select input but normal input, this will allow typing 60 days for examples.

samuveth:
Hey guys, as I mentioned during our last call, we already have a suitable component available on Tune. Check it out here: https://tune-histoire.vercel.app/story/src-components-tuneinputduration-story-vue?variantId=_default

It's simple to adapt for use on SX by applying styling to classes like `tune-input-duration`. Additionally, it would be beneficial to start incorporating Tune components whenever possible. This will allow us to enhance it over time and ensure its compatibility with SX as needed.

src/components/S/IDuration.vue
```diff
@@ -0,0 +1,65 @@
...
+const days = ref(0);
+const hours = ref(0);
+const minutes = ref(0);
+
+watch([days, hours, minutes], () => {
+  const value = (days.value * 24 * 60 + hours.value * 60 + minutes.value) * 60;
+  emit('update:modelValue', value);
+});
+
+watch(
```
samuveth, line 25: Why not use computed with get and set instead of 2 watchers? Example here: https://github.com/snapshot-labs/tune/blob/7e2a1558768f5142701dafb60d181025039c6dfc/src/components/TuneInputDuration.vue#L19

src/components/S/IDuration.vue
```diff
@@ -0,0 +1,65 @@
...
+  },
+  { immediate: true }
+);
+</script>
+
+<template>
+  <div>
+    <div v-text="definition.title" />
+    <div class="flex gap-3">
+      <SBase :definition="{ title: 'Days' }" class="flex-1">
```
samuveth, line 42: When I hover the input the cursor isn't a pointer

src/components/Block/SpaceFormVoting.vue
```diff
@@ -26,18 +26,18 @@ const definition = {
   properties: {
     votingDelay: {
       type: 'number',
-      title: 'Voting delay',
-      examples: ['0']
+      format: 'duration',
```
samuveth, line 29: How will validation work for this duration format? AFAIK the JSON schema "duration" format is expecting a string but we have a number. `For example, P3D expresses a duration of 3 days.`

Source: https://json-schema.org/understanding-json-schema/reference/string.html

Sekhmet: `duration` format is defined for strings, but we use it on number/integer type fields. It's a bit ugly that we have this format conflict though, but I'm not sure if it's something we should change, can't find a better alternative name for it.

src/components/S/IDuration.vue
Sekhmet: I think that using computed here could possibly simplify our watchers (don't think they are that bad anyway), but it will complicate how we handle our individual inputs.

Sekhmet:
@bonustrack should I update this PR to use Tune instead?

bonustrack (approved):
tACK

PR #634, opened 2023-05-10
refactor: make Editor respond to route changes

Sekhmet:
Currently Editor doesn't work with changing routes as it handles some set up at mount time only.
This PR makes this component work with dynamic routes that makes it possible for it to be reused across rerenders instead of being unmounted and remounted as currently (we do this by setting key on router component), but it's wasteful. We should instead make our views handle those changes properly:
- https://router.vuejs.org/guide/advanced/data-fetching.html#fetching-after-navigation

This commit only handles Editor, but there are other views that have the same issue (but are easier to refactor), so will be updated separately.

## Test plan

- Play with editor, nothing should be broken.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -58,7 +58,7 @@ watch(route, () => {
           'translate-x-[72px] lg:translate-x-0': uiStore.sidebarOpen
         }"
       >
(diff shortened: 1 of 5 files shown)


bonustrack (approved):
tACK

PR #635, opened 2023-05-10
refactor: handle route params with reactivity

Sekhmet:
Depends on https://github.com/snapshot-labs/sx-ui/pull/634

This is continuation of Editor refactor, this time removing static handling of route params in other components.

With those changes we can now remove workaround for :key on router to force remounts and instead can reuse when routes change.

## Test plan

- Play with the app, nothing should break.


bonustrack (approved):
utACK

Issue #389, opened 2023-01-26
feat: add ENS domain name resolver


bonustrack:
Currently on SX UI the spaces and profile contain a long address on URLs example:
https://snapshotx.xyz/#/0x07e6e9047eb910f84f7e3b86cea7b1d7779c109c970a39b54379c1f4fa395b28/proposal/70

We should implement a domain name resolver to be able to use a domain from ENS, Lens, Unstoppable domains and Starknet.id example:
https://snapshotx.xyz/#/fabien.eth
https://snapshotx.xyz/#/fabien.lens/proposal/70
https://snapshotx.xyz/#/fabien.crypto/treasury
https://snapshotx.xyz/#/profile/fabien.stark

On the UI we should convert the domain to an address with a network prefix. Lens is on Polygon, Stark on StarkNet and for .eth we should recognize only mainnet domains.

Sekhmet:
We will need to remove dependency on networkID prefix from URL in some places of the app.

Currently we don't really support any of the networks mentioned here (no support for polygon and eth mainnet, Starknet is broken because the contracts now have incompatible API). Does it mean we will need to add support for those networks before we can implement this PR? I could add just the resolvers but they will be pointing to broken URLs and won't be possible to test.

bonustrack:
I imagine for a first step we can try to add support for ENS assuming we use Goerli as default network where fabien.eth is converted to `gor:0x12345...` based on Goerli registry, then later we can switch to mainnet by default for ENS. 

The owner of the NFT can not be the space itself cuz the space contract is not a treasury (can't transfer fund directly) if someone send the NFT to the space contract the space wont be able to transfer it anymore. We can identify the space with the ETH address defined on the ENS profile instead.

<img width="817" alt="image" src="https://user-images.githubusercontent.com/16245250/236790804-6390fcdd-6f49-4b6e-8de6-3cafe3015096.png">

I just created this ENS domain on Goerli: https://app.ens.domains/discopolo.eth
The ENS ETH address is linked to this space: https://snapshotx.xyz/#/gor:0x5acb7858d0441f618fa25cd81dd1b22d1ed90876

When we go to https://snapshotx.xyz/#/discopolo.eth the UI should act like we are on: https://snapshotx.xyz/#/gor:0x5acb7858d0441f618fa25cd81dd1b22d1ed90876

bonustrack:
[...]

PR #637, opened 2023-05-11
feat: add composable based resolver

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/389

This PR implements ENS resolver that uses composable to perform resolving.

I experimented with different approaches, but neither seems super clean.
1. composable approach (this PR) - underlying app uses raw NetworkID/address pair still, conversion is done wherever we accept domain from user. Benefits are that the rest of the app doesn't need to worry about different domain names. Con is that you have to resolve (by using `useResolve` composable) in all places that you expect to see domains.
2. store approach - we could modify our stores to accept IDs instead of NetworkID/address pairs when fetching and reading data (`fetchSpace(id: string)`, `getSpace(id: string)`) and let those stores to perform conversion (resolve on fetch, lookup from `namesStore` when getting values). The issues I've found with this approach are:
    1. quite often we still need to resolve networkId/address at view level so we can't just stick with string ids there (for example some components need Network access, or need to know the actual address of the space).
    2. with above it's unclear what where should we actually initiate resolver to fire - it has to be initiated by `fetchSpace`, because it needs it to work, but if view component also needs networkId/address combo and just uses `namesStore.getName('ethdomain.eth')`  (static helper that reads it from store without actually resolving itself) does it also need to trigger `namesStore.resolveName`? It doesn't have to if we call fetchSpace in the same place (which was the case in all cases I tried), but depending on this staying constant behavior seems wrong.
 
 I think that approach 1. is overall nicer.
 
 Known edgecases/UX issues:
- If you create draft on editor accessed through ENS domain and you select that draft from Draft modal it will navigate you to gor:0x0000 variant.
- Same with favorites - favorites bar only store gor:0x0000 addresses.
- If you try to access draft via URL with ENS domain directly (not from space where ENS was already resolved like http://localhost:8080/#/s1.sekhmet.eth/create/p35t9) it will take a while to load as it needs to resolve name first.
 
 ## Test plan
 
 - http://localhost:8080/#/s1.sekhmet.eth
 - Try to find places where it breaks (editor, editing proposals, drafts, etc.)

### DIFF
--- a/src/helpers/resolver.ts
+++ b/src/helpers/resolver.ts
@@ -1,0 +1,71 @@
+import { getProvider } from '@/helpers/provider';
+import { NetworkID } from '@/types';
+
+const provider = getProvider(5);
+
+type ResolvedName = {
+  networkId: NetworkID;
+  address: string;
+};
+
+function memoize<T extends any[], U>(fn: (...args: T) => U) {
+  const cache = new Map<string, any>();
+
+  return (...args: T): U => {
+    const key = JSON.stringify(args);
+
+    if (cache.has(key)) {
+      return cache.get(key);
+    }
+
+    const result = fn(...args);
+    cache.set(key, result);
+
+    return result;
+  };
+}
+
+function createResolver() {
+  const cache = new Map<string, ResolvedName | null>();
+
+  function resolveStatic(id: string): ResolvedName | null {
+    const parts = id.split(':');
+
+    return {
+      networkId: parts[0] as NetworkID,
+      address: parts[1]
+    };
+  }
+
+  async function resolveEns(id: string): Promise<ResolvedName | null> {
+    const resolvedAddress = await provider.resolveName(id);
+
+    if (!resolvedAddress) {
+      return null;
+    }
+
+    return {
+      networkId: 'gor',
+      address: resolvedAddress.toLocaleLowerCase()
+    };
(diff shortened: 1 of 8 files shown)


src/helpers/resolver.ts
```diff
@@ -0,0 +1,71 @@
...
+    const parts = id.split(':');
+
+    return {
+      networkId: parts[0] as NetworkID,
+      address: parts[1]
+    };
+  }
+
+  async function resolveEns(id: string): Promise<ResolvedName | null> {
+    const resolvedAddress = await provider.resolveName(id);
```
bonustrack, line 41: This is taking 3 requests in serie which is pretty slow, can you try instead to do a call to the function `addr()` to this contract:
```
0xd7a4F6473f32aC2Af804B3686AE8F1932bC35750
```
This is the resolver for `eth` domain on Goerli. 

You will need to pass a hash of the id as `node` like this:
```ts
import { namehash } from '@ethersproject/hash';

const name = namehash('discopolo.eth');
const address = await contract.addr(name);
```

Sekhmet: It's way better. Thanks.

bonustrack (approved):
I'm fine with approach 1 tACK

PR #638, opened 2023-05-15
feat: resolve ENS names for delegates

bonustrack:
This PR add ENS names resolver on the delegates page

image

### DIFF
--- a/src/composables/useDelegates.ts
+++ b/src/composables/useDelegates.ts
@@ -10,6 +10,7 @@ type ApiDelegate = {
 };
 
 type Delegate = ApiDelegate & {
+  name: string | null;
   delegatorsPercentage: number;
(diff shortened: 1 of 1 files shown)


src/composables/useDelegates.ts
```diff
@@ -93,6 +97,7 @@ export function useDelegates(delegationApiUrl: string) {
         (Number(delegate.delegatedVotes) / Number(governanceData.delegatedVotes)) * 100;
 
       return {
+        name: names[delegate.id] || null,
```
Sekhmet, line 100: It should be added in the type as well:
https://github.com/snapshot-labs/sx-ui/blob/5e40f34bbd0931ccef8f22cebf178726e1aa74c2/src/composables/useDelegates.ts#L12-L15

Issue #639, opened 2023-05-16
fix: disable Confirm button for Transaction modal if not configured properly


Sekhmet:
Currently you can put random address in Transaction modal and click confirm.
It shouldn't be possible.

bonustrack:
We should also show an error message if the address is not a contract

PR #641, opened 2023-05-17
feat: disable Confirm button when transaction is invalid

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/639

This PR disabled Confirm button when transaction is invalid:
1. contract address is invalid
2. ABI is not defined (not found or user input doesn't work)
3. method values are invalid

Had to make some adjustments to make it work:
1. Update dirty flag on inputs when value changes from the outside.
2. Make `abiToDefinition` put all inputs to required property.

## Test plan

- Go to create proposal page, select execution -> Transaction.
- Input address that is not contract, error message is shown, button is disabled.
- Input address that is a contract, select method with inputs, do not file them in, button is disabled.
- Correct errors in the form, confirm button is enabled.

### DIFF
--- a/src/components/Modal/Transaction.vue
+++ b/src/components/Modal/Transaction.vue
@@ -89,7 +89,10 @@ const errors = computed(() => {
 });
 const argsErrors = computed(() => validateForm(definition.value, form.args));
 const formValid = computed(
-  () => Object.keys(errors.value).length === 0 && Object.keys(argsErrors.value).length === 0
+  () =>
+    form.abi.length > 0 &&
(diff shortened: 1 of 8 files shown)


bonustrack:
Something weird, if I try to add this address `0x2b5e7ac0e76dcb41d3e506d58c9997d42d133e07` (a space contract) then on the ABI type `[]` it will remove the ABI input.

bonustrack:
Can we show an error if ABI input is not valid?

bonustrack (approved):
tACK

PR #645, opened 2023-05-23
feat: fetch starred spaces

Sekhmet:
To display avatars with ?cb parameter we need to know actual URL of avatar to detect changes of avatars.

Currently it's impossible to compute that parameter as we only have space IDs, we need to fetch Space data from API to be able to compute it.

## Test plan
- Add spaces to favorites, remove it, reorder them - it works as expected.


bonustrack (approved):
tACK

Issue #640, opened 2023-05-16
fix: invalidate Stamp cache when changing avatars


Sekhmet:

Sekhmet:
@bonustrack 

From Notion:
> Maybe we should allow cache to depend on some v query param so we can version avatar changes and fetch latest one?

> We do that on Snapshot, there is a param “cb” that you can add on stamp image URLs, cb stand for cache buster, we use a hash of the avatar image URL for the value. Example: https://cdn.stamp.fyi/space/fabien.eth?s=160&cb=b7fd812c7341f97e

What about the AWS cache on stamp? It doesn't depend on `cb` query, only `type`, `network`, `address`, `w`, `h` and `fallback`. So even with `cb` we can be using old AWS cache?

Sekhmet:
@bonustrack from what I understand we store files on AWS forever so I don't understand how it would ever not use it anymore. That's why I was asking if we have some expiration for AWS, because otherwise it doesn't make sense - I assumed that's the way we wanted it to work for some reason.

Maybe it worked before, because cache was broken (https://github.com/snapshot-labs/stamp/pull/39) - but it would still use cached (original) image from AWS before this PR, it would just keep resizing it on each request.

For example this avatar returns outdated avatar, unless you clear `/clear`
https://stamp.fyi/space-sx/0x65e4329e8c0fba31883b98e2cf3e81d3cdcac780?s=180&cb=20000000 (updated image, but not the latest)
this is the actual image it should return now: https://ipfs.io/ipfs/bafkreih7gpbvmiqgzkzwwhwo2a5ritsezhx4jofvp7pcjdvlf7rv7uss64

If it's not the way we expect it to work then Stamp needs some architectural changes.

bonustrack:
@Sekhmet There is not expiration with AWS, we generate `key1` based on `type`, `network`, `address`, `w`, `h` and `fallback`. In `key1` the `w` and `h` are always 500. This key become a folder and within this folder we store the different resize as cache. When you call `/clear` it will delete the whole folder, so the next time you query it wont use the cache. We also have a browser cache set with `Cache-Control` header which should get reset if you change the `cb`. This is supposed to work this way but there might be some issues if it doesn't react this way.

Sekhmet:
`Cache-Control` only controls how device should cache it (we probably should also use edge-cache so it gets cached by Cloudflare for all users). If we change `cb` parameter it's considered different URL so cache no longer applies.

[...]

PR #646, opened 2023-05-23
feat: add cache bust query parameter for space images

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/640
Depends on: https://github.com/snapshot-labs/sx-ui/pull/645

This won't currently prevent old images from being used still as `cb` is ignored for AWS cache right now (pending https://github.com/snapshot-labs/stamp/pull/55)
You can test just the presence of `cb` parameter on space URLs (user URLs are not currently handled as there are plenty of places where we don't have access to real avatar URL to compute hash so for now it just handles spaces).

## Test plan
- Check different space image URLs.
- They have &cb parameter.

### DIFF
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -26,7 +26,7 @@ const spacesStore = useSpacesStore();
           class="block"
           @click="uiStore.sidebarOpen = false"
         >
-          <Stamp :id="element.id" :size="32" type="space-sx" class="!rounded-[4px]" />
+          <SpaceAvatar :space="element" :size="32" class="!rounded-[4px]" />
         </router-link>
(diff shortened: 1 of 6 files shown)


bonustrack:
The `cb` param is the same for all space for me.

Sekhmet:
If spaces have no avatar they will the same cb as avatar is just `''`. If you look at space with avatar set it will have different value:
http://localhost:8080/#/gor:0xae56be6b3754e5e89de438f6202df07c7aa86845

bonustrack:
Should we remove cb for space without avatar then?

bonustrack (approved):
tACK

PR #647, opened 2023-05-23
fix: hide empty execution block

Sekhmet:
For spaces with no executions we shouldn't show execution block at all - now it just shows empty Execution header.

## Test plan
- Go to editor here http://localhost:8080/#/gor:0xfaa1e9cec71a2b56387aac1b368ef1a852891208


bonustrack (approved):
tACK

PR #648, opened 2023-05-23
feat: add walletconnect goerli RPC node

Sekhmet:
Currently connecting with goerli we get lots of failed requests because we don't have goerli node defined. Also updated mainnet URL.


bonustrack (approved):
utACK

Issue #558, opened 2023-04-19
feat: support Sepolia


bonustrack:
We should add suppport for Sepolia network

Sekhmet:
Currently Sepolia is not supported on Graph's hosted service.

PR #649, opened 2023-05-23
feat: add WalletConnect support with direct transactions

Sekhmet:
This makes it possible to use Safe with WalletConnect to propose/vote with spaces that have Ethereum transaction authenticator supported as well as manage space settings.

Closes: https://github.com/snapshot-labs/sx-ui/issues/605
Depends on https://github.com/snapshot-labs/sx-ui/pull/652
Depends on https://github.com/snapshot-labs/sx.js/pull/194

## Test plan
- Test different spaces:
- Ethereum signature only: http://localhost:8080/#/gor:0x65e4329e8c0fba31883b98e2cf3e81d3cdcac780/settings
- Ethereum transaction only: http://localhost:8080/#/gor:0xfaa1e9cec71a2b56387aac1b368ef1a852891208
- Both: http://localhost:8080/#/gor:0x2b5e7ac0e76dcb41d3e506d58c9997d42d133e07

PR #654, opened 2023-05-29
feat: add sepolia support

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/558
Depends on: https://github.com/snapshot-labs/sx-ui/pull/649

It works, but there seems to be an issue with Graph's file data source where we can't load existing DB entries inside IPFS handlers and we end up with unrecognized executions on spaces.
Contacted Graph about it.


src/networks/evm/index.ts
```diff
@@ -10,19 +10,21 @@ import { NetworkID, Space } from '@/types';
 const METADATA = {
   gor: {
     name: 'Ethereum Goerli',
-    chainId: 5
+    chainId: 5,
+    apiUrl: 'https://api.studio.thegraph.com/query/41343/sekhmet-sx-goerli/version/latest'
```
Sekhmet, line 14: @bonustrack could you share links to our Studio URLs? They have some non-public IDs.

bonustrack: sx-goerli https://api.studio.thegraph.com/query/23545/sx-goerli/version/latest
sx-sepolia https://api.studio.thegraph.com/query/23545/sx-sepolia/version/latest

bonustrack: There isn't any "production deployment" with API key for these, I only see such URL on the mainnet subgraph Studio, probably testnet don't need such keys

bonustrack:
On space creation network tab we should have a hand cursor on the network block, otherwise it works well tACK!

bonustrack (approved):
tACK

PR #664, opened 2023-06-09
feat: add addEthereumChain call for Linea

Sekhmet:
This PR will automatically add Linea to Metamask if it's not already added, without it it will fail with no feedback.

## Test plan
- Start with Metamask with no linea added (you can remove it in settings if it's not active).
- Try voting/proposing.
- You get prompted to add it.
- Try exiting at different stages of the flow (adding network, switching network, confirming transaction).
- Send transaction to vote/propose.
- Nothing breaks.

### DIFF
--- a/src/composables/useActions.ts
+++ b/src/composables/useActions.ts
@@ -25,7 +25,12 @@ export function useActions() {
       try {
         return await fn(...args);
       } catch (e) {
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
tACK

PR #665, opened 2023-06-09
feat: add pagination to proposal votes modal

Sekhmet:
## Test plan

- Change LIMIT to 3, and [modal height to 200px](https://github.com/snapshot-labs/sx-ui/blob/531f749b334e310fc2020c7eb06b7bc429eea7f1/src/components/Ui/Modal.vue#L102)
- Go there: http://localhost:8080/#/gor:0xae56be6b3754e5e89de438f6202df07c7aa86845/proposal/3
- Click votes.
- New votes get loaded in.
- Scroll a bit.
- It triggers another load to fetch the rest.

### DIFF
--- a/src/components/Modal/Votes.vue
+++ b/src/components/Modal/Votes.vue
@@ -23,13 +23,18 @@ const { open } = toRefs(props);
 
 const network = computed(() => getNetwork(props.proposal.network));
 
-watch(open, async () => {
-  if (open.value === false) return;
+function reset() {
+  votes.value = [];
+  loaded.value = false;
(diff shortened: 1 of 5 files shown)


bonustrack:
Looks good! Just some improvements:
- Can we add a top border on the loading to fit with the end of the latest item?
image
- When I load multiple page then close the modal and open it again it will show all the page for few miliseconds then it get reset and start reloading all the pages one by one, would it be possible to either keep the state with all loaded pages or just reset the modal when we open it again?

bonustrack (approved):
Works well tACK

PR #668, opened 2023-06-10
feat: block voting again for same proposal

Sekhmet:
This commit adds pendingVotes in state (it's lost on refresh), so after voting you can't vote for the same proposal again.

This is simple workaround to make UX better. In the future we probably should have better way to handle it where pending vote gets treated same as real ones with a way to handle cases where pending vote actually doesn't execute (transaction failure etc.)

### DIFF
--- a/src/components/Proposal.vue
+++ b/src/components/Proposal.vue
@@ -58,6 +58,7 @@ async function handleVoteClick(choice: Choice) {
         <Vote :proposal="proposal">
           <template #unsupported><div /></template>
           <template #waiting><div /></template>
+          <template #voted-pending><div /></template>
           <template #voted>
             <Results :proposal="proposal" />
           </template>
--- a/src/components/Vote.vue
+++ b/src/components/Vote.vue
@@ -1,10 +1,12 @@
 <script setup lang="ts">
 import { _t } from '@/helpers/utils';
+import { useUiStore } from '@/stores/ui';
 import { getNetwork } from '@/networks';
-import type { Proposal as ProposalType } from '@/types';
+import { Proposal as ProposalType } from '@/types';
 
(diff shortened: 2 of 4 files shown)


bonustrack:
I got this error with estimate gas with this PR for some reason
image
Seem to work on https://snapshotx.xyz

bonustrack:
Hum actually it work sometimes, not sure if it's related to this PR.

Sekhmet:
@bonustrack are you trying to vote on cancelled proposal? Those will show up in UI until we merge cancel PR.

bonustrack:
Oops good catch, I was doing that.

bonustrack:
Can't get a tx onchain for some reason (using EthSig authenticator), tried couple time with the PR or snapshotx.xyz here:
https://goerli.etherscan.io/tx/0xc8f5ff8b332671c752894ee40097549b464b4bdf6f00927efb2f829a5ec63bfd
https://goerli.etherscan.io/tx/0x363f834f0ea8085fa0b521ae58c65cacac371755198f5c8f772459431f477540

Sekhmet:
@bonustrack nonce manager from ethers.js is broken, it probably bumped nonce for failed transactions and is now stuck. We will have to write our own solution for managing nonces probably.

bonustrack:
I've restarted Mana and now it work again. Some related discussion here: https://github.com/ethers-io/ethers.js/issues/972

bonustrack:
Can we display results here after user vote?
image

Sekhmet:
@bonustrack I didn’t do that because those would be inaccurate (your vote won’t be accounted for yet) unless we want to go all the way right away and implement full optimistic updates like described in PR message.

bonustrack (approved):
Ok let's keep it like this for now tACK

Issue #674, opened 2023-06-12
feat: dropdown component


bonustrack:
We should create a dropdown component (can be using floating-ui, similar than tooltips), this component can be used on proposal page here to include "Edit proposal", "Cancel proposal", "View metadata":
<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/d975eccf-baa3-4640-93f9-641c1fb124fe">
Design can be similar than tooltip (using border color as bg)
<img width="897" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/1673727e-f54a-4cb1-8e5b-ba771db9a443">

PR #666, opened 2023-06-09
feat: add proposal cancellation support

Sekhmet:
This PR allows to cancel proposals.

Depends on: https://github.com/snapshot-labs/sx-subgraph/pull/27 (needs to be added to all subgraphs), see test plan how to test before it's deployed.

## Test plan
- Disable all networks but `gor`.
- Change `gor` apiUrl to https://api.studio.thegraph.com/query/41343/sekhmet-sx-goerli/version/latest.
- Create proposal.
- Go to that proposal as space owner.
- You can cancel it.
- Once cancelled it won't show up in proposal list.

PR #675, opened 2023-06-12
feat: add Dropdown

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/674
Depends on: https://github.com/snapshot-labs/sx-ui/pull/666

## Test plan
- Disable linea.
- Create proposal.
- You can open dropdown.

### DIFF
--- a/src/views/Proposal.vue
+++ b/src/views/Proposal.vue
@@ -184,37 +184,46 @@ watch(
             </div>
           </router-link>
           <UiDropdown class="mr-3">
-            <UiDropdownItem v-if="editable" v-slot="{ active }">
-              <button
-                class="flex items-center gap-2"
-                :class="{ 'opacity-80': active }"
-                @click="handleEditClick"
+            <template #button>
+              <IH-dots-vertical class="text-skin-link" />
+            </template>
+            <template #items>
+              <UiDropdownItem v-if="editable" v-slot="{ active }">
+                <button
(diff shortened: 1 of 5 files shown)


src/views/Proposal.vue
```diff
@@ -185,24 +185,39 @@ watch(
...
-              </a>
-            </UiTooltip>
-            <UiTooltip title="View proposal metadata">
-              <a v-if="proposalMetadataUrl" :href="proposalMetadataUrl" target="_blank">
-                <IH-dots-vertical />
+          <UiDropdown class="mr-3">
+            <UiDropdownItem v-if="editable" v-slot="{ active }">
+              <button
+                class="flex items-center gap-2"
+                :class="{ 'opacity-80': active }"
```
bonustrack, line 192: Can we set this style within the Component itself so we don't need to replicate it elsewhere and we got same style across each dropdown?

bonustrack: When "active" is being used? Isn't this similar than "hover"? We can keep it like this if it's to complicated.

Sekhmet: Headless UI supports accessibility so active will be triggered if you focus in with keyboard, if you open popup and start typing part of word from item etc.

bonustrack:
Can you add hand cursor on the 3 dot menu?

bonustrack (approved):
tACK

Issue #619, opened 2023-04-28
style: add border-radius on cover


bonustrack:
We should add this style for cover on profile and space overview pages for small screen.

<img width="499" alt="image" src="https://user-images.githubusercontent.com/16245250/235190902-8b1a8150-9d3d-4bad-9f9d-cda77ccb99a4.png">

<img width="498" alt="image" src="https://user-images.githubusercontent.com/16245250/235191328-01e9cd53-1a43-4995-a385-b92806b941e9.png">

This uses a border radius of 16px.

PR #678, opened 2023-06-14
feat: add Space cover support

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/580
Closes: https://github.com/snapshot-labs/sx-ui/issues/658

Depends on: https://github.com/snapshot-labs/stamp/pull/56
Depends on: https://github.com/pulls

## Test plan

- Use local stamp from PR and change URL in Stamp.vue
- Change goerli API URL to https://api.studio.thegraph.com/query/41343/sekhmet-sx-goerli/version/latest
- Disable all other networks.
- You can see cover photo there.
- You can add cover photos to your own spaces.

## Further comments

We might want to lift size limit on pineapple too, some images will fail to upload now.

PR #679, opened 2023-06-14
feat: add border to cover photo

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/619
Depends on: https://github.com/snapshot-labs/sx-ui/pull/678

## Test plan
- Set up like https://github.com/snapshot-labs/sx-ui/pull/678
- Go to space/user profile.

### DIFF
--- a/src/components/SpaceCover.vue
+++ b/src/components/SpaceCover.vue
@@ -11,5 +11,5 @@ const cb = computed(() =>
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


bonustrack:
Looks cool! 

- Can we make that the space avatar image (without counting the border) start at the same level than the text
image

- Also the end of the cover and start of the avatar should be at the same level on small or large screen for space and users
image

PR #681, opened 2023-06-15
feat: clear pending transactions after 5 minutes

Sekhmet:
## Test plan

- Change txId to `0xe1d29d7207d7fb4b834e4f56c66554b101dfdb7d835809791c01c9193eec3c59` in addPendingTransaction.
- Do some action (Mana looks broken now so check something else).
- Refresh page after 5 minutes, pending transaction is cleared.


bonustrack (approved):
tACK

PR #688, opened 2023-06-16
fix: do not mutate filter for favorite spaces

Sekhmet:
We were mutating filter so spaces were removed for subsequent requests, causing some spaces not to load properly.

## Test plan
- Favorites should load properly.


bonustrack (approved):
tACK

Issue #686, opened 2023-06-16
style: don't show min. end if min is same as max


bonustrack:
On the duration modal if min and max duration are the same we should only show "End".

Before:
<img width="488" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/97744bdc-09df-4865-ade3-6bc39bdcd565">

After:
<img width="484" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/7610413f-3ac0-4c1d-97f7-0537a186eeb8">

PR #689, opened 2023-06-16
feat: merge min end and max end into single step if same

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/686

## Test plan
- Go to space with equal min/max durations and create new proposal (or visit existing if one was created with those settings) - http://localhost:8080/#/gor:0x62f8a028433d811cc0a63c974207276600322062/proposal/8
- Click on date to open timeline.


bonustrack (approved):
tACK

Issue #687, opened 2023-06-16
feat: add filter by status on proposals page


bonustrack:
We should add a button on proposals page to filter proposals by status. Status would be "All", "Pending", "Active", "Closed".
<img width="1455" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/aeffe37e-9a1e-4b37-b07a-2ac49c64c650">

PR #696, opened 2023-06-21
feat: add proposals filter

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/687

## Test plan
- Go to http://localhost:8080/#/gor:0xae56be6b3754e5e89de438f6202df07c7aa86845/proposals
- Test different filters

### DIFF
--- a/src/components/Ui/Dropdown.vue
+++ b/src/components/Ui/Dropdown.vue
@@ -8,7 +8,7 @@ withDefaults(
   }>(),
   {
     gap: '8px',
-    placement: 'left'
+    placement: 'right'
   }
 );
 </script>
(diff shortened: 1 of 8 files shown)


bonustrack:
Can we use the dropdown component for the filters? We should also remove color bullet for "All", and move "Pending" before "Active".

Sekhmet:
> Can we use the dropdown component for the filters?

I used separate one as Select uses different styles than Dropdown. In current design there were too many differences to unify those, but I mostly guessed how we want it to look like, plain Dropdown didn't look good.

If you have some particular style in mind I can implement that but will need some kind of design as reference.

bonustrack:
What's bug you with dropdown design, I tried locally and to me it's looks good.

bonustrack (approved):
tACK

Issue #545, opened 2023-04-03
feat: validate min and maxVotingDuration


zzuziak:
Let's make sure that `minVotingDuration` can't be longer than the `maxVotingDuration`:

<img width="594" alt="Screenshot 2023-04-03 at 14 51 14" src="https://user-images.githubusercontent.com/39556343/229514801-b411229d-825e-4baa-afa8-ac4540ae5a28.png">

PR #700, opened 2023-06-23
feat: validate min max voting duration with each other

Sekhmet:
## Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/545

This PR adds validation for min/max voting duration.

- When creating space it validates max isn't lower than min.
- When editing space it validates each value separately (because they are updated independently).

## Test plan
- Try setting wrong values when deploying space.
- Try setting wrong values when editing space.


bonustrack (approved):
tACK

PR #702, opened 2023-06-23
fix: display tooltips at top of overview buttons

Sekhmet:
## Summary

overflow-hidden was added before that caused it to move to bottom position (to prevent tooltips from being partially cut off). Removing it restores old placement, and it doesn't seem to cause any other issues.


bonustrack (approved):
tACK

PR #703, opened 2023-06-23
chore: handle getNames failure

Sekhmet:
It seems that some addresses cause the call to revert, breaking logging in. For example calling get names with 0xaDa7E6d162C3Ff6f493Cc848f9FE1Bd74dD70A17 will throw.

This change will swallow errors and return empty object if this happens.


bonustrack (approved):
utACK

Issue #705, opened 2023-06-24
style: remove double border on contract call contact modal


bonustrack:
We should remove the doubled border on the contract call execution modal on contacts
<img width="502" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/c92533cf-3489-4b45-aa3b-ee6a6f33b7fa">

PR #706, opened 2023-06-26
fix: prevent double border on Transaction modal

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/705

## Test plan

- Go to editor select transaction execution
- Open contact picker
- There is no double border.


bonustrack (approved):
tACK

Issue #196, opened 2022-07-05
feat: simulate transaction using Tenderly


bonustrack:
When creating a proposal it would be great if we can simulate the transactions to make sure everything will work as expected, we can use Tenderly API for that, here is more info https://docs.tenderly.co/simulations-and-forks/simulation-api
This is similar than what Gnosis Safe released recently https://twitter.com/gnosisSafe/status/1544258597080539137

bonustrack:
Design can look like this
<img width="712" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/76b426c3-aca8-4dd7-a011-08af43dbacc8">

PR #709, opened 2023-06-29
feat: add execution simulation

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/196

This PR adds execution simulation (excluding linea as it's not supported by Tenderly).

## Test plan
- Create proposal draft with Execution.
- Click on button to test if it gets simulated correctly.


bonustrack:
@Sekhmet Can we add a link to the succeed or failed simulation?

bonustrack:
This can be in the notification message

Sekhmet:
> Can we add a link to the succeed or failed simulation?

We execute them as a batch, so with multiple transactions we will have multiple links. Not sure how to display those in that case.

bonustrack:
Good point, let's keep it like this for now then

bonustrack (approved):
tACK

Issue #698, opened 2023-06-22
feat: image upload for proposal content


bonustrack:
Design TBD

PR #711, opened 2023-06-29
feat: add image upload to editor

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/698

This PR adds image upload via drag and drop or clicking bottom bar similar to Github.

### DIFF
--- a/src/components/Markdown.vue
+++ b/src/components/Markdown.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
 import { computed } from 'vue';
 import { Remarkable } from 'remarkable';
+import { getUrl } from '@/helpers/utils';
 
 const props = defineProps<{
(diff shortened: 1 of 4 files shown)


bonustrack:
Awesome! Few suggestions:

- Use ipfs:// URLs instead of the URL with gateway, and replace the images link in preview / proposal content on the fly with the current IPFS gateway setup.
- Remove the footer bar and instead add an image icon in the header bar.
- We could remove "Description" label in the textarea, it's taking space for not much reason there is already label for this field.

bonustrack:
Also can we show an error message if the image upload fail?

bonustrack:
Still not seeing error message when uploading large image

bonustrack (approved):
tACK

Issue #710, opened 2023-06-29
feat: upgrade to WalletConnect v2


bonustrack:
WC v1 is shutdown, we need to move to v2

PR #717, opened 2023-07-03
feat: upgrade to WalletConnect v2

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/710

## Test plan
- Try signing in with WalletConnect.
- Perform some actions on the UI.

### DIFF
--- a/src/helpers/connectors.ts
+++ b/src/helpers/connectors.ts
@@ -20,10 +20,16 @@ export default {
     network: '1',
     icon: walletconnectIcon,
     options: {
-      rpc: {
-        '1': 'https://rpc.snapshotx.xyz/1',
-        '5': 'https://rpc.snapshotx.xyz/5'
-      }
+      projectId: 'e6454bd61aba40b786e866a69bd4c5c6',
+      chains: [5],
+      optionalChains: [1, 11155111],
+      methods: ['eth_sendTransaction', 'eth_signTypedData_v4'],
+      rpcMap: {
+        '1': 'https://rpc.snapshot.org/1',
+        '5': 'https://rpc.snapshot.org/5',
+        '11155111': 'https://rpc.snapshot.org/11155111'
+      },
+      showQrModal: true
     }
   },
   walletlink: {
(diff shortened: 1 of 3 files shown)


src/helpers/connectors.ts
```diff
@@ -20,10 +20,16 @@ export default {
     network: '1',
     icon: walletconnectIcon,
     options: {
-      rpc: {
-        '1': 'https://rpc.snapshotx.xyz/1',
-        '5': 'https://rpc.snapshotx.xyz/5'
-      }
+      projectId: 'e6454bd61aba40b786e866a69bd4c5c6',
```
Sekhmet, line 23: Project ID from https://github.com/snapshot-labs/snapshot/pull/3963

bonustrack:
Hum, doesnt work for me somehow, on MetaMask nothing happen, on Argent it doesn't allow me to connect:
Screenshot_20230703_190733_Argent
Screenshot_20230703_190808_MetaMask

Sekhmet:
@bonustrack Metamask works for me on iOS (same with Zerion).
Argent seems that doesn't work, because we are trying to connect to Goerli and Argent doesn't like it. I set default chainId to Goerli, because otherwise it's impossible to connect to goerli Safe. I'm not sure how we can handle that given we have single app for multiple networks.

bonustrack:
Do you have some Android device to test with?

bonustrack:
I've updated MM on mobile and it seem to work now

src/helpers/connectors.ts
```diff
@@ -20,10 +20,16 @@ export default {
...
+      chains: [5],
+      optionalChains: [1, 11155111],
+      methods: ['eth_sendTransaction', 'eth_signTypedData_v4'],
+      rpcMap: {
+        '1': 'https://rpc.snapshot.org/1',
+        '5': 'https://rpc.snapshot.org/5',
+        '11155111': 'https://rpc.snapshot.org/11155111'
+      },
```
bonustrack, line 31: This seem to be optional, seem to work without it
```suggestion
```

bonustrack (approved):
tACK

Issue #693, opened 2023-06-17
feat: support ENS domain as input on tx builder


bonustrack:
We should support ENS domain in the tx builder for address inputs, domain should be converted to address.

PR #718, opened 2023-07-04
feat: add ENS support for execution forms

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/693

This PR adds async form validation support, `ens-or-address` format and ability to create execution with ENS names.

## Test plan
- Create proposal with execution here: http://localhost:8080/#/gor:0x310d0c40c6d03e0d4189ceb733fc57efab803ea0
- Use ENS domains (created on Goerli) as recipient on Token/NFT execution or in arguments for Contract call.
- You can save/edit/simulate and create proposal with those.
- You can execute that transaction.

### DIFF
--- a/src/components/Modal/SendNft.vue
+++ b/src/components/Modal/SendNft.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import { createSendNftTransaction } from '@/helpers/transactions';
 import { clone } from '@/helpers/utils';
-import { validateForm } from '@/helpers/validation';
+import { getValidator } from '@/helpers/validation';
 
 const DEFAULT_FORM_STATE = {
   to: '',
@@ -11,7 +11,7 @@ const DEFAULT_FORM_STATE = {
 
 const RECIPIENT_DEFINITION = {
   type: 'string',
-  format: 'address',
+  format: 'ens-or-address',
   title: 'Recipient',
   examples: ['Address']
 };
@@ -25,12 +25,25 @@ const props = defineProps({
   initialState: Object
 });
 
+const formValidator = getValidator({
+  $async: true,
+  type: 'object',
+  title: 'TokenTransfer',
+  additionalProperties: false,
+  required: ['to'],
+  properties: {
+    to: RECIPIENT_DEFINITION
+  }
+});
+
 const emit = defineEmits(['add', 'close']);
 
 const searchInput: Ref<HTMLElement | null> = ref(null);
 const showPicker = ref(false);
 const pickerType: Ref<'nft' | 'contact' | null> = ref(null);
 const searchValue = ref('');
+const formValidated = ref(false);
+const formErrors = ref({} as Record<string, any>);
 
 const form: { to: string; nft: string; amount: string | number } = reactive(
   clone(DEFAULT_FORM_STATE)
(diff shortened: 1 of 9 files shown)


src/components/Modal/SendNft.vue
```diff
@@ -11,7 +11,7 @@ const DEFAULT_FORM_STATE = {
 
 const RECIPIENT_DEFINITION = {
   type: 'string',
-  format: 'address',
+  format: 'ens-or-address',
   title: 'Recipient',
   examples: ['Address']
```
bonustrack, line 16:
```suggestion
  examples: ['Address or ENS']
```

src/components/Modal/SendToken.vue
```diff
@@ -16,11 +16,22 @@ const DEFAULT_FORM_STATE = {
 
 const RECIPIENT_DEFINITION = {
   type: 'string',
-  format: 'address',
+  format: 'ens-or-address',
   title: 'Recipient',
   examples: ['Address']
```
bonustrack, line 21:
```suggestion
  examples: ['Address or ENS']
```

bonustrack (commented):
Doesn't work for me on custom contract call execution modal:
image
Error seems like it use address validation no `ens-or-address`.

Also we can add the `ens-or-address` format on "Add contact" modal

Sekhmet:
> Doesn't work for me on custom contract call execution modal:

I haven't added it there intentionally as I thought contracts are unlikely to have ENS domain for it (there is list of things that currently support ENS in PR description).

bonustrack:
It's not common but there is contracts with an ENS domain, it's also likely that space contracts will have their domain too. It would help change space settings.

Sekhmet:
> Also we can add the ens-or-address format on "Add contact" modal

This will require a lot of refactoring across whole app as now we will need to support resolving ENS back to address everywhere we support Contact picker - and there is quite a bit of places outside of execution that need it - unless we convert it to address on save, which is not ideal I think either.

bonustrack:
My idea was to turn it to address on save, isn't how it's done with send token?

bonustrack:
But if thats too much refactoring we can skip it

Sekhmet:
> My idea was to turn it to address on save, isn't how it's done with send token?

It's converted to address where it's needed (part of transaction), but in terms of form data we are storing ENS domain as-is.

With contact we could store both address and raw input if we want (but it might get outdated if someone ever updates ENS domain if we care about it).

bonustrack:
On the modal "Contract call" when I type an ENS domain I can't see the generated or ABI form, here is an example with `discopolo.eth`:
image

bonustrack:
Also on "Contacts" page, there isn't any validation of ENS domains, I can add "blablabla" as address it would work, ideally it should resolve and store the address, or if too much work we can simply remove ENS support on this form.

Sekhmet:
> Also on "Contacts" page, there isn't any validation of ENS domains, I can add "blablabla" as address it would work, ideally it should resolve and store the address, or if too much work we can simply remove ENS support on this form.

I didn't add support for ENS there ([context](https://github.com/snapshot-labs/sx-ui/pull/718#issuecomment-1621647427)). It's missing validation though that I can add (unrelated to this PR though).

bonustrack:
Can we just remove "or ENS" on contact modal field?

Sekhmet:
> On the modal "Contract call" when I type an ENS domain I can't see the generated or ABI form, here is an example with discopolo.eth:

Unrelated, same with putting address directly: `0xae56be6b3754e5e89de438f6202df07c7aa86845`
This address has no functions that we can use, what should we do then? Show message with "No methods found in ABI"? Or ask for different ABI?

bonustrack:
> Unrelated, same with putting address directly: `0xae56be6b3754e5e89de438f6202df07c7aa86845` This address has no functions that we can use, what should we do then? Show message with "No methods found in ABI"? Or ask for different ABI?

This only happen if the contract ABI is known but there isn't any write function? If so we can keep it like this.

bonustrack:
Looks like if I add "0x1" as contact address now it's considered as valid

PR #719, opened 2023-07-25
feat: add v1.0.0 contracts compatibility

Sekhmet:
## Summary

Depends on:  https://github.com/snapshot-labs/sx-subgraph/pull/32
Depends on: https://github.com/snapshot-labs/mana/pull/105

Limitations:
- Timeline component doesn't work as expected as it will need to know conversion from block to future timestamps (or needs to be redesigned to work with blocks).
- There is extra step on app load to fetch latest block number to compare with start/end blocks on proposals - does it need to be handled differently?

## Test plan
- Everything should work except for limitations mentioned above.

### DIFF
--- a/.env
+++ b/.env
@@ -1,5 +1,5 @@
 VITE_ETH_RPC_URL=https://rpc.snapshotx.xyz
-VITE_MANA_URL=https://mana.pizza
+VITE_MANA_URL=http://localhost:3000
 VITE_IPFS_GATEWAY=snapshot.4everland.link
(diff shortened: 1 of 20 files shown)


bonustrack:
tACK

> There is extra step on app load to fetch latest block number to compare with start/end blocks on proposals - does it need to be handled differently?

This sound fine to me.

Can you enable Goerli? The subgraph is deployed there and on Sepolia.

bonustrack (approved):
Looks good, I've deployed the subgraphs for Polygon and Arbitrum here: https://github.com/snapshot-labs/sx-subgraph/pull/33 so later we can add them too.

PR #728, opened 2023-08-16
fix: convert block to timestamp on timeline

bonustrack:
This PR fixes the time in timeline, it convert block number to timestamp using current block.

**Before**
image

**After**
image

### DIFF
--- a/src/components/Modal/Timeline.vue
+++ b/src/components/Modal/Timeline.vue
@@ -23,16 +23,16 @@ const labels = {
 
 const metaStore = useMetaStore();
 
-function getEstimatedTsFromBlock(network, blockNum) {
+function getTsFromBlock(network, blockNum) {
   const networkBlockNum = metaStore.currentBlocks.get(network) || 0;
   const blockDiff = networkBlockNum - blockNum;
 
-  if (blockDiff > 0) return metaStore.currentTs - METADATA[network].blockTime * blockDiff;
-  if (blockDiff < 0) return metaStore.currentTs + METADATA[network].blockTime * Math.abs(blockDiff);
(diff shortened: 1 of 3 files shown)


src/components/Modal/Timeline.vue
```diff
@@ -19,6 +21,17 @@ const labels = {
...
 
+const metaStore = useMetaStore();
+
+function getEstimatedTsFromBlock(network, blockNum) {
+  const networkBlockNum = metaStore.currentBlocks.get(network) || 0;
+  const blockDiff = networkBlockNum - blockNum;
+
+  if (blockDiff > 0) return metaStore.currentTs - METADATA[network].blockTime * blockDiff;
+  if (blockDiff < 0) return metaStore.currentTs + METADATA[network].blockTime * Math.abs(blockDiff);
+  return metaStore.currentTs;
```
Sekhmet, line 32:
```suggestion
  return metaStore.currentTs - METADATA[network].blockTime * blockDiff;
```

Should be the same I think.

src/stores/meta.ts
```diff
@@ -5,6 +5,7 @@ import { enabledNetworks, getNetwork } from '@/networks';
 
 export const useMetaStore = defineStore('meta', () => {
   const loaded = ref(false);
+  const currentTs = ~~(Date.now() / 1e3);
```
Sekhmet, line 8:
```suggestion
  const currentTs = Date.now() / 1e3;
```

src/components/Modal/Timeline.vue
bonustrack: Indeed

PR #730, opened 2023-08-30
feat: load current block at space level

### LINKED ISSUE #729: chore: remove global loading
We should remove the global loading, the loading is only here to load latest block on each chain, instead we should only load latest block once we load a space and only for the network of the space.

bonustrack:
Closes #729 

This PR remove the global loading to get all networks current blocks and instead load current block for a single network when a space or a proposal is loaded. Also this add auto-import for the stores.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,10 +1,6 @@
 <script setup lang="ts">
-import { useUiStore } from '@/stores/ui';
-import { useMetaStore } from '@/stores/meta';
-
 const route = useRoute();
 const uiStore = useUiStore();
-const metaStore = useMetaStore();
 const { modalOpen } = useModal();
 const { userSkin } = useUserSkin();
 const { init, app } = useApp();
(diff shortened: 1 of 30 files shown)


bitcoinfinancier (approved):

src/App.vue
```diff
@@ -68,15 +63,3 @@ watch(route, () => {
...
-.backdrop {
-  position: fixed;
-  top: 0;
-  bottom: 0;
-  left: 0;
-  right: 0;
-  z-index: 99;
-  background: rgba(0, 0, 0, 0.4);
-}
-</style>
```
Sekhmet, line 82: Why was this removed? Without this we won't have backdrop when opening sidebar. Is this intentional?

Sekhmet (commented):
One comment, otherwise looks good to me.

src/App.vue
bonustrack: Good catch, I was assuming it was a duplicate code with the modal.

PR #734, opened 2023-09-06
feat: replace currentBlock with generic current

Sekhmet:
`current` now represents generic current point that can be either block (EVM) or timestamp (starknet) or something else in the future.

By making it generic we can support both types at the same time.

### DIFF
--- a/src/components/Modal/Timeline.vue
+++ b/src/components/Modal/Timeline.vue
@@ -19,7 +19,7 @@ const labels = {
   max_end: 'Max. end'
 };
 
-const { getTsFromBlock } = useMetaStore();
+const { getTsFromCurrent } = useMetaStore();
 
 const states = computed(() => {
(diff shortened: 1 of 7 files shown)


bonustrack:
You can test update proposal here: https://bafybeifcidvuwoa4ry7dzlwie32jsxksxly4fxrrfgoxk2fy74i7vxfsmy.on.fleek.co/#/sep:0x012b261effbf548f2b9a495d50b81a8a7c1dd941/proposal/13 If you login with fabien.eth on https://guest.so

bonustrack:
tACK. I'm now able to see "Edit proposal" in the menu and get on the editor page, but somehow it fail when I try to confirm edition. This is not related to this PR so feel free to merge it, I created an issue here: https://github.com/snapshot-labs/sx-ui/issues/735

PR #736, opened 2023-09-07
feat: add initial sx-starknet support

Sekhmet:
This PR adds initial sx-starknet support that allows you to deploy new spaces, propose and vote. There will be still some UI inconsistencies.

Supported:
- Vanillas
- StarkTx/StarkSig authenticator
- ERC20Votes strategy

Depends on: https://github.com/snapshot-labs/sx-api/pull/181 (link master checkpoint).
Depends on: https://github.com/snapshot-labs/mana/pull/109


bonustrack:
Awesome, I was able to create a space, propose and vote. It would be great to have this before merging:

- For Starknet space creation to set duration it should say "in seconds" not "in blocks"
- Would be great if we can keep the others networks available on space creation like Arbitrum Polygon Goerli etc
- After creating a space indexing the space is stuck for me, I had to manually find the space address
- Explore isn't working for me, got this error:
image

Sekhmet:
> After creating a space indexing the space is stuck for me, I had to manually find the space address

This was fixed in sx.js

> Explore isn't working for me, got this error:

I deployed one space wrong and it was missing something in event. Use latest sx-api with bumped start index. In the future we need to define what happens when we encounter invalid data in events/metadata and how to react (what happens if we receive space metadata that is not JSON or something like that, update, reset, etc.) as right now it's not super consistent and some bugs slip.

.env
```diff
@@ -1,5 +1,5 @@
 VITE_ETH_RPC_URL=https://rpc.snapshotx.xyz
-VITE_MANA_URL=https://mana.pizza
+VITE_MANA_URL=http://localhost:3001
```
bonustrack, line 2: I just need this after merging Mana PR to do my review:
```suggestion
VITE_MANA_URL=https://mana.pizza
```

src/networks/starknet/constants.ts
```diff
@@ -1,89 +1,194 @@
...
+export const PROPOSAL_VALIDATIONS = {
+  '0x38f034f17941669555fca61c43c67a517263aaaab833b26a1ab877a21c0bb6d': 'Voting power'
+};
 
 export const STRATEGIES = {
-  '0x58623786b93d9b6ed1f83cec5c6fa6bea5f399d2795ee56a6123bdd83f5aa48': 'Vanilla',
-  '0xd1b81feff3095ca9517fdfc7427e742ce96f7ca8f3b2664a21b2fba552493b': 'Single slot proof'
+  '0x510d1e6d386a2adcfc6f2a57f80c4c4268baeccbd4a09334e843b17ce9225ee': 'Vanilla',
+  '0x297fb0104d8be6c86f168bf1dcdc28b0625d2b06108c3c46194aa4415f2e2ec': 'Merkle whitelist',
+  '0x619040eb54857252396d0bf337dc7a7f98182fa015c11578201105038106cb7': 'Delegated ERC20 Token'
```
bonustrack, line 46: This should have same name than "OpenZeppelin Votes", actually both "OpenZeppelin Votes" and "Delegated ERC20 Token" are not ideal name, we should change both the EVM and Starknet strategies names to "ERC-20 Votes (EIP-5805)"

bonustrack: Also "ERC-20 Votes (EIP-5805)" should appear before "Delegated Comp Token", and "Delegated Comp Token" should be renamed to "ERC-20 Votes Comp"

bonustrack:
- Creating a space seem to work well, it just a bit slow, we may want to enable optimistic indexing on the API to see how it goes. 
- On space creation for proposal duration it would be great to show selector with days hours minutes, could come with another PR, the same can be done with EVM chain but require more logic to convert to block using blockTime values from: <https://github.com/snapshot-labs/sx-ui/blob/43c945825936953a61f6b545749640ebb2d1e79f/src/networks/evm/index.ts#L17> 
- When I try to create a proposal using StarknetSig authenticator using this space `sn-tn:0x077e73bea0dfa4935e1dd7a3d9f1dae6561b28db0cc6098246300526a93bcbd3` I got this error from Mana:
image

Sekhmet:
> Creating a space seem to work well, it just a bit slow, we may want to enable optimistic indexing on the API to see how it goes.

Enabled there https://github.com/snapshot-labs/sx-api/pull/182

> On space creation for proposal duration it would be great to show selector with days hours minutes, could come with another PR, the same can be done with EVM chain but require more logic to convert to block using blockTime values from:

I made it use IDuration here for Starknet

> We can move Starknet explorer to Starkscan, seem more popular: https://testnet.starkscan.co/

We use starkscan already, only voyager usage was in account picker, but I updated it now too.

> When I try to create a proposal using StarknetSig authenticator using this space sn-tn:0x077e73bea0dfa4935e1dd7a3d9f1dae6561b28db0cc6098246300526a93bcbd3 I got this error from Mana:

I can't reproduce this error right now, in case of Mana errors we need to check Mana logs.

bonustrack:
Ok creating a proposal or cast vote is working for me now, issue was that my account was not upgraded to Cairo 1.

bonustrack (approved):
tACK

Issue #725, opened 2023-08-07
bug: create a space timelock show error directly


bonustrack:
When I create a space and select Timelock execution the modal open with an error on delay:
<img width="511" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/ef493540-55c4-48dc-8be1-3fcca9eff28b">
The error should not be visible directly.

PR #743, opened 2023-09-20
feat: add dirty flag to IDuration

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/725

Now error state won't be shown unless user interacted with input


bonustrack (approved):
tACK

Issue #722, opened 2023-08-05
bug: cover is lost when updating profile


bonustrack:
When I update a space profile the cover got removed.

PR #744, opened 2023-09-20
fix: keep previous cover when updating space metadata

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/722


bonustrack (approved):
tACK

PR #745, opened 2023-09-20
fix: use voting power voting strategies when proposing

Sekhmet:
## Test plan

- Propose there: http://localhost:8080/#/sn-tn:0x04b1953ac6bbe7a162d1afca0065bc82ba707501acbc01774cc1cf59721da6df


bonustrack (approved):
tACK

Issue #723, opened 2023-08-05
feature(treasury): custom base token


bonustrack:
For a space on Polygon the base token should be MATIC, currently it show as ETH
<img width="340" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/8290cc87-ef4d-4b28-9277-0d6c9f155fdb">
See example:
https://snapshotx.xyz/#/matic:0xa98d476118af1a2c9eab803db9688e33cf68435f/treasury

PR #746, opened 2023-09-20
feat: customize base token for each EVM network

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/723

Depends on: https://github.com/snapshot-labs/stamp/pull/88

## Test plan
- Run local stamp for logo support: https://github.com/snapshot-labs/stamp/pull/88
- Check http://localhost:8080/#/matic:0xa98d476118af1a2c9eab803db9688e33cf68435f/treasury

## Screenshots

image

### DIFF
--- a/src/composables/useTreasury.ts
+++ b/src/composables/useTreasury.ts
@@ -19,6 +19,7 @@ export function useTreasury(spaceRef: Ref<NullableSpace>) {
     if (!chainId || !wallet) return null;
 
     return {
+      shortName: networkId,
       network: chainId,
       wallet
     };
--- a/src/composables/useTreasury.ts
+++ b/src/composables/useTreasury.ts
@@ -19,7 +19,7 @@ export function useTreasury(spaceRef: Ref<NullableSpace>) {
     if (!chainId || !wallet) return null;
 
     return {
-      shortName: networkId,
+      networkId,
       network: chainId,
       wallet
     };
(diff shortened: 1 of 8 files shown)


bonustrack:
This wont work for DAOs on Arbitrum as the base token is ETH, now it show Arbitrum as token:
`http://localhost:8080/#/arb1:0x2fb231a3ac7c40888d2e6773f5600851b0c20ce4/treasury`. Also it would be better to move the logic to resolve token images to Stamp.

Sekhmet:
@bonustrack moved it to Stamp.

src/views/Space/Treasury.vue
```diff
@@ -175,7 +175,12 @@ watchEffect(() => {
             class="mx-4 py-3 border-b flex"
           >
             <div class="flex-auto flex items-center min-w-0">
-              <Stamp :id="asset.contractAddress" type="token" :size="32" />
+              <Stamp
+                :id="`eip155:${treasury.network}:${asset.contractAddress}`"
```
bonustrack, line 179: @Sekhmet Wouldn't it be simpler to just use network short name instead of eip155 name?

bonustrack (approved):
tACK

bonustrack (changes requested):
Actually just realized but we also have token in the "Send token" modal, would be great to do same change.

src/composables/useTreasury.ts
```diff
@@ -19,6 +19,7 @@ export function useTreasury(spaceRef: Ref<NullableSpace>) {
     if (!chainId || !wallet) return null;
 
     return {
+      shortName: networkId,
```
bonustrack, line 22: We could keep `networkId` here to be consistent with others files 
We could 
```suggestion
     networkId,
```

bonustrack (approved):

Issue #750, opened 2023-09-20
feat(sx-evm): duration form for voting duration inputs


bonustrack:
On the duration input for SX-EVM we should use the duration form with day / hours / seconds. We can estimate the number of block based on the average block duration here: https://github.com/snapshot-labs/sx-ui/blob/53aba99e9426b4905ab842f0f4d259d09d714056/src/networks/evm/index.ts#L17

PR #756, opened 2023-09-22
feat: use IDuration input for sx-evm with block estimations

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/750

This PR adds IDuration input for sx-evm voting delay and voting periods with blocks being estimated from duration.

## Test plan
- Deploy new EVM space.
- Go to deployed space settings, values are formatted as durations there.
- Try editing one of the settings, everything works as expected.


bonustrack (approved):
tACK

Issue #749, opened 2023-09-20
chore(sx-sn): remove vanilla authenticator


bonustrack:
We should remove Vanilla authenticator for SX Starknet

PR #758, opened 2023-09-22
chore(sx-sn): remove vanilla authenticator

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/749


bonustrack (approved):
uTACK

Issue #622, opened 2023-05-01
feat: support whitelist voting strategy


bonustrack:
TBD

PR #759, opened 2023-09-27
feat: add whitelist support with simplified UI

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/622

This PR adds support for whitelist strategies with simplified UI (textarea where you put addresses in specified format), no custom UI as of now.

## Test plan

- Add Starknet/EVM whitelist strategy to space.
- Try propose/vote.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@snapshot-labs/lock": "^0.2.0",
     "@snapshot-labs/pineapple": "^1.1.0",
     "@snapshot-labs/prettier-config": "^0.1.0-beta.7",
-    "@snapshot-labs/sx": "^0.1.0-beta.39",
+    "@snapshot-labs/sx": "^0.1.0-beta.40",
     "@vueuse/core": "^10.4.1",
     "@walletconnect/client": "^1.8.0",
     "ajv": "^8.12.0",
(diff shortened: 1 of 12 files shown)


bonustrack:
The UI looks good enough that's what I was thinking. I just worry for the usercase where there is like 200k addresses in the whitelist. I don't think that list should be stored in the API directly.

bonustrack:
For the strategy description we can have this:

> A strategy that defines a list of addresses each with designated voting powers, using a Merkle tree for verification of these allocations.

Sekhmet:
@bonustrack should be ready for review

bonustrack:
This is in standby we should use the merkle tree whitelist strategy for both EVM and Starknet, but the voting strategy on EVM is not deployed yet it's here: https://github.com/snapshot-labs/sx-evm/pull/241 .

Sekhmet:
@bonustrack can we proceed with this PR anyway? I need those changes to implement strategies editing (and if we keep this PR open as a base it will delay it until we are can get whitelist deployed, implemented on sx.js, released etc.)
If we don't want current sx-evm whitelist we can just remove it from possible editor strategies.

bonustrack:
Ok sure

bonustrack (approved):
utACK, would be nice to just disable EVM whitelist for now (the one without merkle tree)

Issue #712, opened 2023-07-01
chore: update dependencies


bonustrack:
This can be used https://github.com/antfu/taze to update dependencies, there is some code changes to do to keep same dimensions with Tailwind or format with Vue.js

PR #761, opened 2023-09-28
chore: upgrade deps

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/712

`starknet` is not upgraded to satisfy compatibility with sx.js


bonustrack:
Can we update GitHub action to fix the test?

bonustrack:
We could move choices.json to constants.ts

bonustrack:
So looks like there is also this file that we need to remove:
https://github.com/snapshot-labs/sx-ui/blob/master/src/helpers/execution.json
This may fix the issue with test / build GitHub action

Sekhmet:
We need to update eslint-config.

bonustrack (approved):
tACK

Issue #732, opened 2023-09-03
feat: assets price on treasury and send token modal


bonustrack:
We should load token price from Coingecko API and show them on the "Treasury" page and "Send token" modal.

PR #763, opened 2023-09-29
feat: add treasury rates via CoinGecko

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/732

## Test plan
- Go to MATIC/ARB space: http://localhost:8080/#/matic:0x4d95a8be4f1d24d50cc0d7b12f5576fa4bbd892b/treasury
- Values are shown.

## Screenshots

image

### DIFF
--- a/src/components/Modal/SendToken.vue
+++ b/src/components/Modal/SendToken.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import { formatUnits } from '@ethersproject/units';
+import { METADATA_BY_CHAIN_ID } from '@/composables/useBalances';
 import { createSendTokenTransaction } from '@/helpers/transactions';
 import { ETH_CONTRACT } from '@/helpers/constants';
 import { clone } from '@/helpers/utils';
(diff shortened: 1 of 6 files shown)


bonustrack:
I found some issue when you switch between space, the native token in send token modal show as ETH instead of MATIC see the flow:
https://www.loom.com/share/0438056d41ff4ed4b23a38455004203e?sid=cd905631-4d56-485b-a22c-02b69bffb76d

bonustrack:
Also:
- Can you change color of the asset value from white to text-skin-text?
image

- And can you force a maximum of 6 decimals when we use dollar amount to convert to token amount here:
image

bonustrack:
Also for the 24h percentage increase can you round the number to 2 decimals?

bonustrack:
Can we also make that the name of the token or the amount + symbol are always shown in 1 line instead of 2 if content is too small? We can use text ellipsis for this
Screenshot_20230930_111104_Chrome

bonustrack:
I still can see the bug with the native token changing from ETH to MATIC when I open Send token modal

bonustrack:
Ok works well now, sorry one last thing, can we show only 2 decimals for the amount in $ here?
image

bonustrack (approved):
tACK

Issue #724, opened 2023-08-05
feat: votes page


bonustrack:
We should add a menu in the proposal page to link to a page "Votes", this page would include the list of all the votes with:

- Voter name or address
- Choice
- Voting power with amount and symbol
- Voting power in % of total vp from votes in the proposal
- Cast date
- Link to transaction

Some requirements:
- The page should include a top banner with the proposal name and space.
- There should be a way to sort by voting power, choice or date, asc and desc
- We should be able to filter results by choice

Here is a simple example on how it can look like (with incomplete info):

<img width="1467" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/c98cb1b0-bd35-441e-9e60-f9103cffa8a3">

I imagine a double line would be better on the list to pass all the information, similar than delegates:
<img width="902" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/e58b0575-f8b9-4676-b973-29730ca71bfd">
(https://snapshotx.xyz/#/sep:0x0f4504ccd7f98368a0f94cde1ade1b624635ffe2/delegates)

We should keep the menu on the right side the same as on the proposal page, ideally without duplicate code, this could be implemented as a base view similar than "Space.vue".

Page route should be in this format: 
https://snapshotx.xyz/#/matic:0xa98d476118af1a2c9eab803db9688e33cf68435f/proposal/3/votes

Issue #607, opened 2023-04-27
feat: add link to the space from the proposal page


bonustrack:
Design TBD

Sekhmet:
@bonustrack can I use the UI from this issue for this?
https://github.com/snapshot-labs/sx-ui/issues/724

![image](https://github.com/snapshot-labs/sx-ui/assets/1968722/e362a0cf-a145-417e-abd7-d576ac7b148d)

bonustrack:
Yes

PR #765, opened 2023-10-02
feat: add votes page

Sekhmet:
Closes: https://github.com/snapshot-labs/sx-ui/issues/724
Closes: https://github.com/snapshot-labs/sx-ui/issues/607

## Test plan
- Use different sorting/filters.

## Screenshots

image

### DIFF
--- a/src/components/Ui/Select.vue
+++ b/src/components/Ui/Select.vue
@@ -1,23 +1,24 @@
-<script setup lang="ts">
-type Item = {
-  key: string;
+<script setup lang="ts" generic="T extends string | number, U extends readonly Item<T>[]">
+export type Item<T extends string | number> = {
+  key: T;
   label: string;
   indicator?: string;
 };
 
 const props = defineProps<{
-  modelValue: string;
+  modelValue: U[number]['key'];
   title: string;
-  items: Item[];
+  items: U;
   gap?: string;
   placement?: 'left' | 'right';
 }>();
 
 const emit = defineEmits<{
-  (e: 'update:modelValue', value: string): void;
+  (e: 'update:modelValue', value: U[number]['key']): void;
 }>();
 
 const currentItem = computed(() => props.items.find(item => item.key === props.modelValue));
+const items = computed(() => props.items);
 </script>
 
 <template>
--- a/src/components/Ui/Select.vue
+++ b/src/components/Ui/Select.vue
@@ -34,9 +34,7 @@ const items = computed(() => props.items);
           <div
             v-if="currentItem.indicator"
             class="w-[8px] h-[8px] rounded-full"
-            :class="{
-              [currentItem.indicator]: true
-            }"
+            :class="currentItem.indicator"
           />
           {{ currentItem.label }}
         </template>
@@ -49,13 +47,7 @@ const items = computed(() => props.items);
           :class="{ 'opacity-80': active, 'opacity-40': disabled }"
           @click="() => emit('update:modelValue', item.key)"
         >
-          <div
-            v-if="item.indicator"
-            class="w-[8px] h-[8px] rounded-full"
-            :class="{
-              [item.indicator]: true
-            }"
-          />
+          <div v-if="item.indicator" class="w-[8px] h-[8px] rounded-full" :class="item.indicator" />
           {{ item.label }}
         </button>
       </UiDropdownItem>
--- a/src/components/Ui/Select.vue
+++ b/src/components/Ui/Select.vue
@@ -7,6 +7,7 @@ type Item = {
 
 const props = defineProps<{
   modelValue: string;
+  title: string;
   items: Item[];
   gap?: string;
   placement?: 'left' | 'right';
@@ -26,7 +27,7 @@ const currentItem = computed(() => props.items.find(item => item.key === props.m
         class="flex items-center gap-2 relative rounded-full leading-[100%] border button px-[16px] min-w-[76px] h-[42px] top-1 outline-0 text-skin-link bg-skin-bg"
       >
         <div class="absolute top-[-10px] bg-skin-bg px-1 left-[12px] text-[16px] text-skin-text">
-          Status
+          {{ title }}
         </div>
         <template v-if="currentItem">
           <div
--- a/src/components/Ui/Select.vue
+++ b/src/components/Ui/Select.vue
@@ -24,21 +24,23 @@ const items = computed(() => props.items);
 <template>
   <UiDropdown :gap="gap" :placement="placement">
     <template #button>
-      <button
-        class="flex items-center gap-2 relative rounded-full leading-[100%] border button px-[16px] min-w-[76px] h-[42px] top-1 outline-0 text-skin-link bg-skin-bg"
-      >
-        <div class="absolute top-[-10px] bg-skin-bg px-1 left-[12px] text-[16px] text-skin-text">
-          {{ title }}
-        </div>
-        <template v-if="currentItem">
-          <div
-            v-if="currentItem.indicator"
-            class="w-[8px] h-[8px] rounded-full"
-            :class="currentItem.indicator"
-          />
-          {{ currentItem.label }}
-        </template>
-      </button>
+      <slot name="button">
+        <button
+          class="flex items-center gap-2 relative rounded-full leading-[100%] border button px-[16px] min-w-[76px] h-[42px] top-1 outline-0 text-skin-link bg-skin-bg"
+        >
+          <div class="absolute top-[-10px] bg-skin-bg px-1 left-[12px] text-[16px] text-skin-text">
+            {{ title }}
+          </div>
+          <template v-if="currentItem">
+            <div
+              v-if="currentItem.indicator"
+              class="w-[8px] h-[8px] rounded-full"
+              :class="currentItem.indicator"
+            />
+            {{ currentItem.label }}
+          </template>
+        </button>
+      </slot>
     </template>
     <template #items>
       <UiDropdownItem v-for="item in items" :key="item.key" v-slot="{ active, disabled }">
--- a/src/networks/common/graphqlApi/index.ts
+++ b/src/networks/common/graphqlApi/index.ts
@@ -105,15 +105,33 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
   return {
     loadProposalVotes: async (
       proposal: Proposal,
-      { limit, skip = 0 }: PaginationOpts
+      { limit, skip = 0 }: PaginationOpts,
+      filter: 'any' | 'for' | 'against' | 'abstain' = 'any',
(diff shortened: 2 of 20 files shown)


bonustrack:
Layout doesn't looks great atm (my fault) could we move to something like this for the overview page:

image

- Using the "Overview | Votes" nav as first element of the page
- Keep title of the proposal how it was previously
- Move author and space name under title of the proposal

For the votes page I suggest to:

- Add label for column, and sorting for Date and Voting power. If the sorting is active the label and icon should be white. If the sorting is not active we shouldn't show the icon.
- Add 3 dots icon in the end of the line, this open a dropdown with a link "View on block explorer"
- Use only 1 line per item (we can remove the % of vp for now)

image

I created this proposal with a bunch of vote to test:
http://localhost:8080/#/gor:0x21e0a186e708094859c064b656037ebbbd61f57b/proposal/2/votes

Also can we merge this to #760 and continue with a single PR?

src/views/Proposal/Votes.vue
```diff
@@ -0,0 +1,218 @@
...
+        <col class="w-[25%] lg:w-[20%]" />
+        <col class="w-[25%] lg:w-[20%]" />
+        <col class="w-[60px] lg:w-[20%]" />
+        <col class="w-[0px] lg:w-[100px]" />
+      </colgroup>
+      <thead>
+        <tr class="border-b">
+          <th class="text-left pl-4 eyebrow text-skin-text pb-2">Votes</th>
+          <th class="eyebrow text-skin-text">
+            <div
```
samuveth, line 120: For accessibility this and the others should be a button I think.

src/views/Proposal/Votes.vue
```diff
@@ -0,0 +1,218 @@
...
+            </div>
+          </th>
+          <th />
+        </tr>
+      </thead>
+      <td v-if="!loaded" colspan="5">
+        <UiLoading class="p-4 block text-center" />
+      </td>
+      <template v-else>
+        <div v-if="votes.length === 0" class="p-4 text-center col-span-full">
```
samuveth, line 153: The positioning of the message here seems off
image

src/views/Proposal/Votes.vue
```diff
@@ -0,0 +1,218 @@
...
+          <th class="eyebrow text-skin-text">
+            <div
+              class="flex justify-end items-center"
+              :class="{
+                'text-skin-link': sortBy === 'created-desc' || sortBy === 'created-asc'
+              }"
+              @click="handleSortChange('created')"
+            >
+              <span>Date</span>
+              <IH-arrow-sm-down v-if="sortBy === 'created-desc'" />
```
samuveth, line 128: Should we hide arrow when filter for the column is not selected?

src/networks/common/graphqlApi/index.ts
```diff
@@ -105,15 +106,33 @@ export function createApi(uri: string, networkId: NetworkID): NetworkApi {
   return {
     loadProposalVotes: async (
```
samuveth, line 107: I'm a bit confused, why is this query in networks folder?

src/views/Proposal/Votes.vue
Sekhmet: I followed the design from [there](https://github.com/snapshot-labs/sx-ui/pull/765#issuecomment-1746845239) which shows icon on multiple comments at the same time (inactive are just grayed out).

Not sure if hiding it might make sense as it won't be clear if that column supports sorting or not (for example Choice is not sortable and doesn't have those arrows).

src/networks/common/graphqlApi/index.ts
Sekhmet: This query or all queries overall?

`networks` directory contains interfaces (both network APIs, contract calls, constants, etc.) for interacting with supported networks (right now `starknet` and `evm`) in unified format. Those have separate actions and constants (due to differences between those), but the network API is shared and identical on both, so we have `common` directory there for things that are common between multiple networks, otherwise they would be in `network/starknet/api` or something like this. I think it makes sense to store those queries in networks directory, because if we had single network, or there was no common APIs between multiple networks those queries would still live in `networks` directory.

src/views/Proposal/Votes.vue
samuveth: I see, there is a comment there by Fabien though:
> If the sorting is not active we shouldn't show the icon.

src/networks/common/graphqlApi/index.ts
samuveth: Makes sense thanks.

src/views/Proposal/Votes.vue
Sekhmet: Oh I see, didn't notice it. Thanks.

samuveth:
By the way, why was the back button removed even though the issue #760 is closed and links to this PR?

Sekhmet:
@samuveth it's no longer a back button as it didn't fit the design, it's now [clickable Space name under proposal title](https://github.com/snapshot-labs/sx-ui/pull/765#issuecomment-1746845239) that can take you back.

src/components/Ui/Select.vue
```diff
@@ -1,22 +1,24 @@
-<script setup lang="ts">
-type Item = {
-  key: string;
+<script setup lang="ts" generic="T extends string | number, U extends readonly Item<T>[]">
```
samuveth, line 1: This seems hard to read in a single line, how about this:

```
type ItemKeyType = string | number;
type ItemArray<T extends ItemKeyType> = readonly Item<T>[];

<script setup lang="ts" generic="T extends ItemKeyType, U extends ItemArray<T>">
```

src/components/Ui/Select.vue
```diff
@@ -1,22 +1,24 @@
...
   placement?: 'left' | 'right';
 }>();
 
 const emit = defineEmits<{
-  (e: 'update:modelValue', value: string): void;
+  (e: 'update:modelValue', value: U[number]['key']): void;
 }>();
 
 const currentItem = computed(() => props.items.find(item => item.key === props.modelValue));
+const items = computed(() => props.items);
```
samuveth, line 21: This doesn't seem to be needed. We can simply use `props.items` or `items` in template

src/components/Ui/Select.vue
```diff
@@ -26,7 +28,7 @@ const currentItem = computed(() => props.items.find(item => item.key === props.m
         class="flex items-center gap-2 relative rounded-full leading-[100%] border button px-[16px] min-w-[76px] h-[42px] top-1 outline-0 text-skin-link bg-skin-bg"
       >
         <div class="absolute top-[-10px] bg-skin-bg px-1 left-[12px] text-[16px] text-skin-text">
-          Status
+          {{ title }}
         </div>
         <template v-if="currentItem">
           <div
```
samuveth, line 34: Any reason to write is like this:

```
:class="{
  [currentItem.indicator]: true
}"
```

instead of:

```
:class="currentItem.indicator"
```

same for line 53.

samuveth (approved):
tACK

Left a few more comments for consideration

src/components/Ui/Select.vue
Sekhmet: I tried this and to me it seems harder to read, because we need to put those helper types below first line so to read you now need to jump back and forth multiple times to understand. In the end there is not much change in the first line, but we add two extra types.

src/components/Ui/Select.vue
Sekhmet: Without this TS types in template get crazy for some reason, looks like some edgecase/bug in Vue when using generics.

```bash
yarn run v1.22.19
$ vue-tsc --noEmit
src/components/Ui/Select.vue:45:56 - error TS2339: Property 'key' does not exist on type 'NonNullable<[{ type: PropType<U>; required: true; }] extends [Prop<infer V, infer D>] ? unknown extends V ? IfAny<V, V, D> : V : { type: PropType<U>; required: true; }>[keyof NonNullable<...>]'.

45       <UiDropdownItem v-for="item in items" :key="item.key" v-slot="{ active, disabled }">
                                                          ~~~

src/components/Ui/Select.vue:49:56 - error TS2339: Property 'key' does not exist on type 'NonNullable<[{ type: PropType<U>; required: true; }] extends [Prop<infer V, infer D>] ? unknown extends V ? IfAny<V, V, D> : V : { type: PropType<U>; required: true; }>[keyof NonNullable<...>]'.

49           @click="() => emit('update:modelValue', item.key)"
                                                          ~~~

src/components/Ui/Select.vue:52:24 - error TS2339: Property 'indicator' does not exist on type 'NonNullable<[{ type: PropType<U>; required: true; }] extends [Prop<infer V, infer D>] ? unknown extends V ? IfAny<V, V, D> : V : { type: PropType<U>; required: true; }>[keyof NonNullable<...>]'.

52             v-if="item.indicator"
                          ~~~~~~~~~

src/components/Ui/Select.vue:55:21 - error TS2339: Property 'indicator' does not exist on type 'NonNullable<[{ type: PropType<U>; required: true; }] extends [Prop<infer V, infer D>] ? unknown extends V ? IfAny<V, V, D> : V : { type: PropType<U>; required: true; }>[keyof NonNullable<...>]'.

55               [item.indicator]: true
                       ~~~~~~~~~

src/components/Ui/Select.vue:58:19 - error TS2339: Property 'label' does not exist on type 'NonNullable<[{ type: PropType<U>; required: true; }] extends [Prop<infer V, infer D>] ? unknown extends V ? IfAny<V, V, D> : V : { type: PropType<U>; required: true; }>[keyof NonNullable<...>]'.

58           {{ item.label }}
                     ~~~~~


Found 5 errors in the same file, starting at: src/components/Ui/Select.vue:45
```

bonustrack:
Sorry for the back and forth but here is the final design for the votes page: https://www.figma.com/proto/Q7y6LYRDtdJrMPYN0qkYKX/Snapshot%3A-votes-page?type=design&node-id=565-20912&t=H1UHRN7XsoT0zaUU-1&scaling=scale-down-width&page-id=419%3A12100&mode=design
You will need to click on top left icon to see comments, I added few of them to give more details

src/components/Ui/Select.vue
samuveth: Ok I see what you mean.

bonustrack:
Looks great! Just some few details:
- "Voting Power" should be "Voting power"
- Against and abstain icons should be same as in the app with the cross and arrow (this was misleading in my Figma I didn't had the icons so I used approve icon)
- The choice icons should be aligned to "Choice" column label and all align to left in the column.
- Ideally the whole column "Date", "Choice" and "Voting power" should be clickable

bonustrack:
Also the choice circle icons should be same height and width than on the overview or proposal page, 40px size, not 48px.

bonustrack:
Also we lost border on modals:
image
Can we have them back?

bonustrack (approved):
tACK

PR #772, opened 2023-10-05
fix: Use auto-imports to resolve Icons outside of template

samuveth:
This isn't as elegant as I initially hoped, and we might want to consider this instead https://www.npmjs.com/package/@iconify/tailwind?activeTab=readme

Do you think it is stable enough for us to use it @Sekhmet?

### DIFF
--- a/src/components/Nav.vue
+++ b/src/components/Nav.vue
@@ -1,16 +1,6 @@
 <script lang="ts" setup>
 import { compareAddresses } from '@/helpers/utils';
 
-// TODO: need to import all icons https://github.com/antfu/unplugin-icons/issues/5
-// move to this when stable to avoid imports https://www.npmjs.com/package/@iconify/tailwind
-import IHGlobeAlt from '~icons/heroicons-outline/globe-alt';
-import IHNewspaper from '~icons/heroicons-outline/newspaper';
-import IHCash from '~icons/heroicons-outline/cash';
-import IHLightningBolt from '~icons/heroicons-outline/lightning-bolt';
-import IHCog from '~icons/heroicons-outline/cog';
-import IHUsers from '~icons/heroicons-outline/users';
-import IHStop from '~icons/heroicons-outline/stop';
-
 const route = useRoute();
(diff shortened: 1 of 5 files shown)


Sekhmet:
> This isn't as elegant as I initially hoped

What exactly isn't elegant in this approach?

Personally coming from non-Vue background I'm not big fan of making everything autoimported (at least until IDE support is improved) as it feels a bit magical and sometimes results in weird conflicts, but I understand it's Vue-way. To me it seems as elegant as autoimports tend to be. You mean that Tailwind style-based icons would be more elegant?

samuveth:
Yeah, I think the auto imports have their limitations, and this is pushing it.

The mismatch in casing and syntax between auto imports and iconify names annoys me. It also feels like a lot to type out. 

With the tailwind based approach, I think we could have a component like this to make it more readable.

Icon.vue
```html
<span :class="`icon-[${name}]`" />
```

And use it like this
```html
<Icon name="mdi-light-home" />
```

Sekhmet:
> With the tailwind based approach, I think we could have a component like this to make it more readable.

Would it work? Generally TailwindCSS only adds classes for styles that you use statically (at least in a sense that transiler can figure out what is used). With string interpolation it might not work (at least it doesn't for regular styles), and opting-in into all icons might be wasteful.

samuveth:
Oh right, probably wouldn't work.

samuveth:
Closing this because importing seems more straight forward and readable.

I also tried `@iconify/tailwind` package which didn't work at all.

samuveth:
@Sekhmet is that comment still relevant now, or can I remove it?

```
// TODO: need to import all icons https://github.com/antfu/unplugin-icons/issues/5
// move to this when stable to avoid imports https://www.npmjs.com/package/@iconify/tailwind
```

Sekhmet:
@samuveth not sure as it was suggested by John, I don't think so, but I'm biased because I l'm okay with explicit imports :D You can remove it if you want.

PR #773, opened 2023-10-06
Create PULL_REQUEST_TEMPLATE.md

samuveth:


Sekhmet (commented):
More of a feedback on the template itself, not really sx-ui specific (I understand we want to consistent templates across repos).

1. Generally there will be single issue to close in each PR (sometimes none). I feel like separate section for it, at the top is too much. Generally it's not that important to highlight it so much.
2. There is issue link and changes list, but not really a place for general motive for the changes and/or explanation for approach taken - maybe "Additional notes or considerations", but it's all the way at the bottom and I feel like it's downplaying the role of such summary.
3. To solve first two "issues" I usually tend to have **Summary** section that could look like this:
```md
### Summary

Closes:

(description of the change)

(could have list of changes from Changes section too).
```
4. Is self-review checklist supposed to be included in final PR? If not we can put HTML comment around it so it won't be visible.

samuveth:
Thanks for suggestions. Made some changes

samuveth:
We could keep the checklist visible to act as a guide for reviewers, letting them know what the author has already considered.

Sekhmet:
> We could keep the checklist visible to act as a guide for reviewers, letting them know what the author has already considered.

I wasn't sure about the intent of that list, before I commented I took a look at one of recent PRs on `snapshot-labs/snapshot` and it had this section removed so I assumed it was for PR author only.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,20 @@
+### Summary
+
+<!-- _*(Concludes of related issues, a description or list of the changes and the motivation behind them)*_ -->
```
Sekhmet, line 3:
```suggestion
<!-- _*(Consists of related issues, a description or list of the changes and the motivation behind them)*_ -->
```

Is Concludes correct in this context? If it is then we might use some more common word to be clearer anyway, or reword it and remove the prefix entirely so it reads "Related issues, a description or list of the changes and the motivation behind them".

btw. what does `_*` & `*_` change there?

samuveth: Nothing, I forgot it from before.

Sekhmet (approved):
utACK

PR #774, opened 2023-10-06
Create ISSUE_TEMPLATE/BUG_REPORT.yml and ISSUE_TEMPLATE/FEATURE_REQUEST.yml

samuveth:
You can still create an issue without the template
image

These are the same templates we have on snapshot repo. You can test it there


Sekhmet (approved):
utACK

Issue #771, opened 2023-10-04
style: bug with flash message width


bonustrack:
There is an issue with the style of the flash message, it use a full width instead of having a dynamic width:
<img width="1469" alt="image" src="https://github.com/snapshot-labs/sx-ui/assets/16245250/dc366d7a-8050-43d9-ae8b-b8141c411ff1">
This is most likely due to the update of Tailwind package.

PR #776, opened 2023-10-10
fix: Notification width

samuveth:
### Summary

Closes: #771

The issues was that somehow `w-fit` wasn't taking effect. Fixes it buy adding `!` and I also increased the gap to 16px. And fixed a typo

Preview:
image

### How to test

1. Make action that triggers Notification


Sekhmet (approved):
utACK

Issue #768, opened 2023-10-04
feat: move to Mixpanel


bonustrack:
We should replace Google Analytics with Mixpanel with geolocation disabled see more info here: https://docs.mixpanel.com/docs/tracking/how-tos/privacy-friendly-tracking#disabling-geolocation
Here is the project token `47b43858cf6fe9376f8deddda52cbaaf`

Sekhmet:
@samuveth 
https://github.com/snapshot-labs/sx-ui/blob/be07543e2216110150124d7ae435c1f9bc46c75d/src/main.ts#L28-L44

PR #777, opened 2023-10-10
feat: Add mixpanel tracking

samuveth:
### Summary

Not sure what the best approach is for Vue 3. There is very little downloads for `vue-mixpanel` and it looks very Vue 2. I figured I could create a composable instead.

Closes: #768

### How to test

1. Go to space create page
2. Check on Mixpanel dashboard if tracking is working

image


src/composables/useMixpanel.ts
```diff
@@ -0,0 +1,7 @@
+import mixpanel from 'mixpanel-browser';
+
+export function useMixpanel() {
+  mixpanel.init('47b43858cf6fe9376f8deddda52cbaaf');
```
bonustrack, line 4: This should be a config var, also there should be config to not track IP

bonustrack:
We should remove Google Analytics, it's setup here: https://github.com/snapshot-labs/sx-ui/blob/master/src/main.ts#L28-L44 
Also I'm not sure about your approach, we don't want to add code in each of the view to track page load, there should be a single setup that take in consideration all the route (similar than GA now).

samuveth:
Waiting for @bonustrack to answer DMs on this. @Sekhmet he seems to believe there is a way to use Mixpanel globally like GA, but all I see from docs or `vue-mixpanel` is that tracking has to be added per route/component. Any ideas?

samuveth:
IMO we should first get an extensive list of things we want to track and then implement tracking on an individual basis. Also worth a read is https://docs.mixpanel.com/docs/getting-started/what-should-I-track

Sekhmet:
> That would only track which routes the user visits, we still wouldn't see on the daskboard if a user voted for example

We can do both, can't we? And tracking routes is what we want right now I believe as that's what this PR does (although manually) and that's what Google Analytics did before.

Sekhmet:
@bonustrack I fixed env file and re-added automatic route tracking as we had with GA. I tested it with my own Mixpanel account.

Not that Brave blocks Mixpanel by default, so we will need to disable protection for it if you use it (likely in uBlock Origin too).

src/router.ts
```diff
@@ -65,4 +67,12 @@ const router = createRouter({
   }
 });
 
+router.afterEach(to => {
+  mixpanel.track('Page View', {
```
bonustrack, line 71: Shouldn't we use the default function `mixpanel.track_pageview` for page views? See: https://docs.mixpanel.com/docs/tracking-methods/sdks/javascript#tracking-page-views

bonustrack: There seem to be a "built-in" event for page view, in my dashboard it show the data to 0, while if I track all event I can see some metric for "Page View" custom event

Sekhmet: > The track_pageview option currently does not auto-track page views in single-page applications. For single-page applications or other manual page view tracking, Mixpanel offers the standard page view event through mixpanel.track_pageview(). This call can be inserted into any hooks or listeners for your framework of choice.

It says auto-tracking doesn't work for SPAs.

I changed it to use `track_pageview`, I left `page_name` and `page_path` as they don't have default replacement (we have full URL, so maybe `page_path` is redundant a bit).

bonustrack (approved):
tACK

src/router.ts
bonustrack: Seem to work for me

Issue #767, opened 2023-10-04
chore: hide "Your voting power" and "Cast your vote" sections if proposal ended


bonustrack:
On the proposal page, we should hide "Your voting power" and "Cast your vote" sections once the proposal ended (with max end)

samuveth:
@Sekhmet any reason you unassigned this issue? PR is already open

Sekhmet:
Sorry, it was unassigned when PR was submitted so I wanted to assign it to you and muscle memory took over and I assigned myself (and later you), and instead of removing myself I removed you by mistake.

PR #779, opened 2023-10-10
fix: Hide voting power and vote section if ended

samuveth:
### Summary

Closes: #767

image

### How to test

1. Go to a proposal that has ended or is canceled 
2. See sidebar is missing vote and voting power section


Sekhmet:
The layout seems to be different:
image

Top gap was used to be smaller before. Now Results is lower than "Cast your vote" section was.

samuveth:
@Sekhmet on the issue is says `with max_end` but on this PR I'm using `has_ended` and `canceled`, is there any difference?

Sekhmet (approved):
tACK

Issue #780, opened 2023-10-11
bug: Closed proposal showing in active


samuveth:
When I first open a space from the Explore page, I see proposals that are actually closed in the `Active` section. But when I reload the space page, it shows them correctly in the `Closed` section.

PR #787, opened 2023-10-19
fix: fetchBlock on loading proposals/spaces

Sekhmet:
### Summary

Because we made calling fetchBlock from global to on-demand there is risk that UI doesn't call it when it's needed (it happened in case of visiting space from explore page where it has space right away.

We avoid this problem by making requesting latest block responsibility of store actions (it won't fetch new block if it's already fetched).

Regression from: https://github.com/snapshot-labs/sx-ui/pull/730

Closes: https://github.com/snapshot-labs/sx-ui/issues/780

### How to test

1. Go to http://localhost:8080/#/explore
2. Click on Shadow DAO
3. Proposals in overview page show as closed proposals.


samuveth (approved):
utACK

Issue #790, opened 2023-10-21
feat: delegate sorting toolbar


bonustrack:
On the delegates page we should add a toolbar similar than on the proposal votes page, to sort delegate by number of delegators or voting power asc and desc. We can also remove the "delegator(s)" from the rows content.

PR #792, opened 2023-10-23
feat: add toolbar to Delegates page

Sekhmet:
### Summary

Closes: [#](https://github.com/snapshot-labs/sx-ui/issues/790)

### How to test

1. Go to http://localhost:8080/#/arb1:0x2fb231a3ac7c40888d2e6773f5600851b0c20ce4/delegates
2. Try sorting by different values.

### Screenshots

image

### DIFF
--- a/src/composables/useDelegates.ts
+++ b/src/composables/useDelegates.ts
@@ -24,8 +24,8 @@ type Governance = {
 const DELEGATES_LIMIT = 40;
 
 const DELEGATES_QUERY = gql`
-  query ($first: Int!, $skip: Int!) {
-    delegates(first: $first, skip: $skip, orderBy: delegatedVotes, orderDirection: desc) {
+  query ($first: Int!, $skip: Int!, $orderBy: Delegate_orderBy!, $orderDirection: OrderDirection!) {
+    delegates(first: $first, skip: $skip, orderBy: $orderBy, orderDirection: $orderDirection) {
       id
       delegatedVotes
       delegatedVotesRaw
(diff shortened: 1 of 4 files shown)


bonustrack:
Looks great, there is just some weird results when query for delegators asc, address 0x0000... is returned, and lot of delegatee without voting power. Can we force a filter where delegator > 0?

src/composables/useDelegates.ts
```diff
@@ -24,8 +24,14 @@ type Governance = {
...
 const DELEGATES_QUERY = gql`
-  query ($first: Int!, $skip: Int!) {
-    delegates(first: $first, skip: $skip, orderBy: delegatedVotes, orderDirection: desc) {
+  query ($first: Int!, $skip: Int!, $orderBy: Delegate_orderBy!, $orderDirection: OrderDirection!) {
+    delegates(
+      first: $first
+      skip: $skip
+      orderBy: $orderBy
+      orderDirection: $orderDirection
+      where: { tokenHoldersRepresentedAmount_gte: 0 }
```
bonustrack, line 33: I think it's better this way:
```suggestion
      where: { tokenHoldersRepresentedAmount_gt: 0 }
```

bonustrack:
Just realized the vertical alignment on the toolbar is off by 1 or 2px
image
Can you fix this here and on the votes pages too?

bonustrack:
Also toolbar should never appear over the topnav, currently if you scroll fast to bottom, the toolbar will overlap with topnav
image

Sekhmet:
> Delegator column should have text aligned to right (similar than others columns with numbers)

I started with this, but I couldn't find a way for it to make sense unless entire column is aligned to right (including column name).

Otherwise it's impossible to make it align to left with column title and first number and right aligned to first number for second number, because those numbers have different lengths.

bonustrack:
I imagine everything aligned to right, column name included.

Sekhmet:
> Just realized the vertical alignment on the toolbar is off by 1 or 2px

Not sure what you mean? Elements are not aligned to each other or all of them are not aligned in the middle?

I tested alignment between each other (looking at `e` character and they are aligned:|
image

Left side:
image

Right side:
image

Also overall alignment seems to be in the middle (not super accurate because of aliasing):
image

bonustrack:
@Sekhmet I mean for the overall alignment, text looks 1px too much on bottom

Sekhmet:
Updated it, but now it looks wrong to me the other way :D

bonustrack:
Another issue, the 1px gap on mobile is back

Screenshot_20231024_155243_Chrome

bonustrack (approved):
tACK

PR #794, opened 2023-10-30
feat: Add histoire for component stories

samuveth:
### Summary

Add Histoire for component stories. The `@esbuild-plugins/node-globals-polyfill` was causing an error with Histoire but it didn't seem needed anymore so I removed it.

### How to test

1. Run `yarn story:dev`


samuveth:
I think this could be useful for larger projects that have a lot of new components. Instead of making one big PR, we could create smaller PRs for each component first. That way, they can be demoed and tested with Histoire before being integrated in the UI.

Sekhmet:
CI links to deployment doesn't work for me. It asks me to sign in (I don't know what I'm supposed to sign in with). Using my own account I get 404 after signing in.

bonustrack:
Invited you @Sekhmet

Sekhmet:
@samuveth can you rebase and see if it still works as expected?

samuveth:
@Sekhmet done, and it works for me in local. We can handle the `histoire.snapshotx.org` deployment later I think

Sekhmet (approved):
tACK

PR #801, opened 2023-11-03
fix: do not reset networksMap when fetching explore page

Sekhmet:
### Summary

We don't fetch all networks for explore page (snapshot.org spaces are not fetched, but we have to keep previous networksMap entries so it doesn't get lost).

### How to test

1. Go to http://localhost:8080/#/s:arbitrumfoundation.eth and favorite this space.
2. Go to http://localhost:8080/#/explore
3. Click on Arbitrum space from favorites list.
4. It loads.

### DIFF
--- a/src/composables/useSpaces.ts
+++ b/src/composables/useSpaces.ts
@@ -108,25 +108,28 @@ export function useSpaces() {
       })
     );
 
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
tACK

PR #802, opened 2023-11-03
feat: display proposal state

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/pitches/issues/51

Changes:
- Until we get real integration of Central icon System it's just embedding SVGs like this. I didn't want to add those as separate icons yet.
- Not showing `#PROPOSAL_NUMBER` for read only spaces as they don't have numerical IDs.
- Pending uses box with minus sign, active uses empty box, passed/executed uses box with plus sign, rejected uses box with cross sign (not all were defined).
- Replaced `has_started`, `executed`, `has_ended` flags for unified `state` prop.

### How to test

1. Go to few spaces, for example
    1. http://localhost:8080/#/s-tn:radiantcapital.eth
    2. http://localhost:8080/#/s:arbitrumfoundation.eth
    3. http://localhost:8080/#/sep:0xea4322cc9be9ae0c9400984209c858031b1c4438
2. Overview only shows single list of proposals.
3. You can see all proposal types (pending/active/passed/rejected/executed)

### Screenshots

image

image


bonustrack:
- For Snapshot proposals we can use the first 7 char of the proposal id (similar than recent change I did on proposal page for title)
- User address should use text-skin-text color.
- There is some proposals here: http://localhost:8080/#/s:arbitrumfoundation.eth/proposals that didn't pass but still showing "Passed" icon instead of "Rejected".
- Which variant did you used from Central Icon? It feel the border are too light, is there a variant with ticker border?
- When proposal title take 2 lines the icon should stay at same level than if there were 1 line, (not centered based on full height):
image
Example from here: http://localhost:8080/#/s:ens.eth/proposals

Sekhmet:
> There is some proposals here: http://localhost:8080/#/s:arbitrumfoundation.eth/proposals that didn't pass but still showing "Passed" icon instead of "Rejected".

I ended up implementing it as "passed" meaning that quorum was met and "rejected" as quorum was not met.

I changed it to have following rules for **closed** proposals:
1. If  quorum is not met: `rejected`.
2. If quorum is met and it's using non-basic voting: `passed`.
3. If quorum is met and votes for > votes against: `passed`.
4. Otherwise `rejected`.

Not sure if it makes sense.

> Which variant did you used from Central Icon? It feel the border are too light, is there a variant with ticker border?

I used this one and icons from below: https://www.figma.com/file/GIS1iuHQiwnim0xLUJ0mbF/centraliconsystemv1.15?type=design&node-id=5480-1507&mode=dev

There are no variants with thicker border it seems.

bonustrack:
Thanks, sorry I should have tested before but these icons and colors are too confusing IMO, I'd like to propose some final changes using Heroicons, here are suggestions:

- Use icon `IS-clock` for `pending` with color `text-gray-400`
- Use icon `IS-status-online` for `active` with color `text-green` (currently the green is doesn't match with others on UI)
- Use icon `IS-check-circle` for `passed` with color `text-skin-link`
- Use icon `IS-play` for `executed` with color `text-purple-500`
- Use icon `IS-x-circle` for `rejected` with color `text-red`

I cons should be 17px size, it would look like this:
image
image

Others suggestions:

- Reduce line height on proposal preview title with  `class="leading-6 my-1"`
- Remove `0x` from the id of Snapshot proposals and show only 5 chars (ideally on proposal page too)

- Use transparent bg for the icons approve reject abstain on proposals or votes pages:
image

Sekhmet:
Just fyi, there is unrelated issue with proposal votes loading indefinitely. I will submit a PR with a fix.

bonustrack (approved):
tACK

PR #807, opened 2023-11-05
chore: Replace lock with wagmi

### LINKED ISSUE #124: Add back agentx connector

### LINKED ISSUE #125: Replace lock with wagmi

### LINKED ISSUE #805: Fix wagmi reconnect

samwhosung:
### Summary

Closes: https://github.com/snapshot-labs/sx-monorepo/issues/125
Closes: https://github.com/snapshot-labs/sx-ui/issues/805
Closes: https://github.com/snapshot-labs/sx-monorepo/issues/124

### How to test

1. Try connecting with all connectors, including safe on safe.global 
2. Try various actions like voting

### To-Do

- [x] Track further todos here https://github.com/snapshot-labs/pitches/issues/32


src/App.vue
```diff
@@ -47,13 +43,13 @@ watch(isSwiping, () => {
 </script>
 
 <template>
+  <div id="modal" />
```
samuveth, line 46: After removing the redundant `app.init` there was an issue with the teleport inside `Topnav` rendering before `<div id="modal" />` which was causing an error

src/helpers/wagmiConfig.ts
```diff
@@ -0,0 +1,41 @@
+import { createConfig, mainnet } from 'use-wagmi';
+import { createPublicClient, http } from 'viem';
+import { InjectedConnector } from 'use-wagmi/connectors/injected';
+import { CoinbaseWalletConnector } from 'use-wagmi/connectors/coinbaseWallet';
+import { WalletConnectConnector } from 'use-wagmi/connectors/walletConnect';
+import { SafeConnector } from 'use-wagmi/connectors/safe';
+
+const config = createConfig({
+  // TODO: Enable after onConnect is fixed https://github.com/unicape/use-wagmi/issues/82
+  // autoConnect: true,
```
samuveth, line 10: I've disabled autoConnect (reconnect) until the issues from the comment is fixed

src/networks/types.ts
```diff
@@ -18,7 +18,7 @@ export type SpacesFilter = {
   controller?: string;
   id_in?: string[];
 };
-export type Connector = 'argentx' | 'injected' | 'walletconnect' | 'walletlink' | 'gnosis';
+export type Connector = 'argentx' | 'injected' | 'walletConnect' | 'coinbaseWallet' | 'safe';
```
samuveth, line 21: Since we are using wagmi now these ids have changed

package.json
```diff
@@ -37,12 +37,10 @@
     "@floating-ui/vue": "^1.0.2",
     "@headlessui/vue": "^1.7.16",
     "@snapshot-labs/eslint-config-vue": "^0.1.0-beta.13",
-    "@snapshot-labs/lock": "^0.2.0",
     "@snapshot-labs/pineapple": "^1.1.0",
     "@snapshot-labs/prettier-config": "^0.1.0-beta.7",
     "@snapshot-labs/sx": "^0.1.0-beta.44",
     "@vueuse/core": "^10.4.1",
-    "@walletconnect/client": "^1.8.0",
```
samuveth, line 45: This seemed redundant

src/components/VotingPowerIndicator.vue
```diff
@@ -43,7 +42,7 @@ function handleModalOpen() {
     >
       <UiTooltip title="Your voting power">
         <UiButton
-          v-if="web3.account && !(evmNetworks.includes(networkId) && web3.type === 'argentx')"
+          v-if="web3Account"
```
samuveth, line 45: Will add this back once we have agentx again

bonustrack:
In the modals it seem like Calibre font is not working can you reproduce that? Also we will need Starknet connector if we want merge this.

samuveth:
@bonustrack you said in DM "I would try integrate first without Argent"?

Fixed the modal text

Sekhmet:
@samuveth we won't be able to merge this without Starknet support because it will cause 50% of the app to be non-functional.

src/components/Modal/Account.vue
```diff
@@ -1,70 +1,85 @@
...
           </UiButton>
-        </a>
+        </template>
       </div>
     </div>
     <div v-else>
-      <div v-if="$auth.isAuthenticated.value" class="m-4 space-y-2">
-        <a :href="explorerUrl(web3.network.key, web3.account)" target="_blank" class="block">
+      <div v-if="isConnected" class="m-4 space-y-2">
+        <a :href="explorerUrl(chain.id, web3Account)" target="_blank" class="block">
```
samuveth, line 79: @Sekhmet how does `explorerUrl` work for `argentx` if it's not in the networks.json?

Sekhmet: It has special case: https://github.com/snapshot-labs/sx-ui/blob/c271cf63de0b37b5adfac86f799f1ad422a5e777/src/helpers/utils.ts#L97

vite.config.ts
```diff
@@ -53,7 +54,12 @@ export default defineConfig({
   optimizeDeps: {
     include: ['@snapshot-labs/sx'],
     esbuildOptions: {
-      target
+      target,
+      plugins: [
+        GlobalPolyFill({
```
samuveth, line 59: Without this Coinbase wallet fails to connect

src/components/Modal/Account.vue
samuveth: @Sekhmet yes but `networks[network].explorer` will still be undefined

samuveth:
@Sekhmet added back the ArgentX/Starknet connector. Please have a look

bonustrack:
@samuveth When I open the deployment here: https://bafybeicwfiiggbhy66b3ztq6ed3xkassephuuzcdo3zzmi4jsbhty5yru4.on.fleek.co/#/ it directly try to log me, I see connect wallet button loading. Can you reproduce?

Also:
- If someone cancel login we shouldn't show the error message "User rejected wallet connection..."
- I can't see my ENS name fabien.eth when I log in, was this removed?
- Ideally Starknet connector should be lazy loaded

samuveth:
> @samuveth When I open the deployment here: https://bafybeicwfiiggbhy66b3ztq6ed3xkassephuuzcdo3zzmi4jsbhty5yru4.on.fleek.co/#/ it directly try to log me, I see connect wallet button loading. Can you reproduce?
> 
> Also:
> 
> * If someone cancel login we shouldn't show the error message "User rejected wallet connection..."
> * I can't see my ENS name fabien.eth when I log in, was this removed?
> * Ideally Starknet connector should be lazy loaded

- It doesn't try to log you in, it's checking what connectors are available in your browser and will only show the ones that are available in the modal. e.g try disabling the ArgentX plugin and see it will no longer show it in the list.
- To not show that error we have to handle that specific error for each wallet or hide all errors? I'd rather show it incase the error is one that is relevant.
- Ens is fixed
- Lazy loading is fixed

bonustrack:
> It doesn't try to log you in, it's checking what connectors are available in your browser and will only show the ones that are available in the modal. e.g try disabling the ArgentX plugin and see it will no longer show it in the list.

Even if it's not trying to connect me we shouldn't make the button loading, this is confusing UX.

> To not show that error we have to handle that specific error for each wallet or hide all errors? I'd rather show it incase the error is one that is relevant.

Hide all errors.

Sekhmet (changes requested):
Issue 1: Connect with braavos, refresh page (modal shows up, you can see you are logged in behind the modal), refresh page again - no longer signed in.
Issue 2: Connect with braavos, refresh page (model shows up, you can see you are logged in behind the modal), close modal, you are still logged in - why was the modal opened if you are logged in?
Issue 3: I can't deploy space on Starknet (or probably anything else, looks like web3 is null).

src/main.ts
```diff
@@ -1,9 +1,8 @@
-import { createPinia } from 'pinia';
-import { LockPlugin } from '@snapshot-labs/lock/plugins/vue3';
-import options from '@/helpers/auth';
 import App from '@/App.vue';
 import router from '@/router';
-import '@/helpers/auth';
+import { createPinia } from 'pinia';
+import { UseWagmiPlugin } from 'use-wagmi';
```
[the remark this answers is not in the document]
samuveth: So `pinia` isn't external. Not sure I get the logic of this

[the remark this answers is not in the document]
samuveth: Oh I put `pinia` below, forgot about that

src/App.vue
```diff
@@ -47,13 +43,8 @@ watch(isSwiping, () => {
...
 
 <template>
-  <div
-    ref="el"
-    :class="{ [skin]: true, 'overflow-hidden': scrollDisabled }"
-    class="font-serif text-base min-h-screen bg-skin-bg text-skin-text antialiased"
-  >
-    <UiLoading v-if="app.loading || !app.init" class="overlay big" />
-    <div v-else class="pb-6 flex">
+  <div ref="el" :class="{ [skin]: true, 'overflow-hidden': scrollDisabled }" class="min-h-screen">
```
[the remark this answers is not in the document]
samuveth: You need UiLoading with useApp? @bonustrack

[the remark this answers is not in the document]
samuveth: What for?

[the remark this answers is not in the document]
bonustrack: To load some metadata for discussions. We will need the loading later when we introduce whitelabel too to get domains list.

bonustrack:
Publish a proposal, cast a vote, edit settings doesn't work for me it return this error:
```
index-c9534916.js:1 TypeError: Cannot read properties of null (reading 'provider')
```
You can use this space to test, it use vanilla (ticket) strategy `sep:0x012b261effbf548f2b9a495d50b81a8a7c1dd941`. I haven't tested with Starknet.

Support for EIP-6963 seam to work well, I've tried using both MetaMask and Rabbi chrome extension and it works. It would be great if we can support wallet logo or show a default logo instead of always MetaMask, we have a list of all wallet and image here: https://github.com/snapshot-labs/lock/blob/master/src/utils.ts Zerion should be added in the list.

image

bonustrack:
I see also there is some increase in bundle size, biggest dependency is Safe SDK, is possible to lazy load it?
https://bafybeigiabs43nayawrgpogf4dohcvgp3ujc53siaote5tfofdwtxkepua.on.fleek.co/stats.html 5.86mb
https://testnet.snapshotx.xyz/stats.html 4.64mb

samuveth:
> I see also there is some increase in bundle size, biggest dependency is Safe SDK, is possible to lazy load it? https://bafybeigiabs43nayawrgpogf4dohcvgp3ujc53siaote5tfofdwtxkepua.on.fleek.co/stats.html 5.86mb https://testnet.snapshotx.xyz/stats.html 4.64mb

Weird I see 8.43mb and 8.92mb. 

Don't see any way to lazy load it, also the SDK is used to detect the safe wallet which is needed to hide/show the connector and auto-connect when used as a safe app.

samuveth:
Made a fix and tested with Metamask, WC, Coinbase. Also tried with ArgentX and got weird network error that isn't related I think, @Sekhmet can you help test argent?

Sekhmet:
> Made a fix and tested with Metamask, WC, Coinbase. Also tried with ArgentX and got weird network error that isn't related I think, @Sekhmet can you help test argent?

Can you rebase on latest master as on this branch we updated contracts and fixed Starknet related issue with sequencer being deprecated? This alone might fix it.

bonustrack:
Can we also remove loading on connect wallet when we open the app?

bonustrack:
I see also there is a lot of JS file being loaded by default compared to current version of SX:

With your PR:
image

Without:
image

Do you know what are these? Maybe something is wrong with lazy loading

samuveth:
> Can we also remove loading on connect wallet when we open the app?

It's connected to isConnecting. I suggest we do like on `hey.xyz`, we remove the loading from "Connect wallet" button completely and do loading in the modal when connecting and close modal after it's connected. We also won't need the loading for reconnect anymore because wagmi.cache will help instantly reconnect without delay

samuveth:
We can reduce the bundle size by replacing etherjs with viem, which is smaller and is used by wagmi already.

bonustrack:
Hum this is problematic if providers aren't lazy loaded.

samuveth:
Seems to be possible in v2, which isn't out of alpha yet I think.

samuveth:
I'm currently talking to wagmi if we can fix lazy loading connectors with v1 or v2. Putting this PR on standby for now

samuveth:
Seems like `use-wagmi` updated to `wagmi-v2` three weeks ago https://github.com/unicape/use-wagmi/commit/73daf2693953fdf9c3a93833769fa6bcc82e06b1

I'll still be busy with boost but could pick this up afterwards

PR #808, opened 2023-11-07
chore: remove node-globals-polyfill

Sekhmet:
### Summary

This is no longer needed.

### How to test

1. App still works (in dev mode and after build).


bonustrack (approved):
utACK

PR #810, opened 2023-11-08
fix: remove subtree option for BlockInfiniteScroller

Sekhmet:
### Summary

BlockInfiniteScroller used to wrap entire table but it wouldn't trigger endReached event because table itself didn't have any direct children added to it, so subtree property was added, but it caused different issue as now table's last element is always visible on screen (tbody) so it triggers endReached event all the time.

subtree property is not good solution and we should always wrap just the elements we want to load.

### How to test

1. Go to http://localhost:8080/#/s:arbitrumfoundation.eth/proposal/0xc040de9c6a85dee5ca15de10691165bf5595245ffac922cbef61e45199522345/votes
2. Check for network requests - no requests being fired all the time.
3. Scroll down, it will fetch new votes as needed.

### DIFF
--- a/src/views/Proposal/Votes.vue
+++ b/src/views/Proposal/Votes.vue
@@ -90,84 +90,85 @@ watch([sortBy, choiceFilter], () => {
 </script>
 
 <template>
-  <BlockInfiniteScroller
-    :subtree="true"
-    :loading-more="loadingMore"
-    @end-reached="handleEndReached"
-  >
-    <table class="text-left w-full table-fixed">
-      <colgroup>
-        <col class="w-[50%] lg:w-[40%]" />
-        <col class="w-[25%] lg:w-[20%]" />
-        <col class="w-[25%] lg:w-[20%]" />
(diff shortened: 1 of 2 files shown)


src/views/Proposal/Votes.vue
```diff
@@ -90,84 +90,85 @@ watch([sortBy, choiceFilter], () => {
...
-      </colgroup>
-      <thead
-        class="bg-skin-bg sticky top-[112px] lg:top-[113px] z-40 after:border-b after:absolute after:w-full"
-      >
-        <tr class="bg-skin-border/10">
-          <th class="pl-4 font-medium">
-            <span class="relative bottom-[1px]">Voter</span>
-          </th>
-          <th class="hidden lg:table-cell">
+  <table class="text-left w-full table-fixed">
```
Sekhmet, line 93: Messy diff because it moved `BlockInfiniteScroller` down the tree.
You can disable whitespace to get cleaner diff:
image

bonustrack (approved):
tACK

PR #811, opened 2023-11-08
fix: do not show read only networks on space creation form

Sekhmet:
### Summary

Filter out read only networks from create space form.

Also created constans for Snapshot network. 

### How to test

1. Go to http://localhost:8080/#/create
2. Name space, go next.
3. Snapshot network is not visible.


bonustrack (approved):
tACK

PR #820, opened 2023-11-25
style: enhance proposal view and tabs

### LINKED ISSUE #814: feat: turn url to link on proposal content
Here is an example of proposal where url is not clickable: https://snapshotx.xyz/#/s:hydraventures.eth/proposal/0xe9639b3e6da6fe83fc3383ab43f1c49d58294806e6ed084568c2e4c83c59f9a5

### LINKED ISSUE #818: feat: show vote count number on tab
We should show the number of vote on the "Votes" tab using similar design than GitHub:
image
image

The counter should not be visible if count is 0.

We should also show a bottom border using link color on the active tab:
image

bonustrack:
Closes #814 
Closes #818 

- Also this PR add markdown support for `backticks` you can try here: http://localhost:8080/#/s:balancer.eth/proposal/0x141051b2fe6f18e322539c0e54b2fb0d9872b6096f2c3fd71fa3bf27e5c7b8f8
- Increase width for proposal overview and editor to 660px
- Update loading layout on proposal votes page to be consistent with others loading
- Add border bottom on active tabs
- Truncate proposal preview title

### DIFF
--- a/src/components/Proposal.vue
+++ b/src/components/Proposal.vue
@@ -32,7 +32,7 @@ async function handleVoteClick(choice: Choice) {
               space: `${route.params.id}`
             }
           }"
-          class="space-x-2 flex max-w-[700px]"
+          class="space-x-2 flex"
         >
           <ProposalStatusIcon width="17" height="17" :state="proposal.state" class="top-[7.5px]" />
           <h3
--- a/src/components/Proposal.vue
+++ b/src/components/Proposal.vue
@@ -23,7 +23,7 @@ async function handleVoteClick(choice: Choice) {
 <template>
   <div>
     <div class="border-b mx-4 py-2.5 flex">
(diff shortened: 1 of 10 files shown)


Sekhmet (approved):
Looks good to me, but truncating the title like this looks weird to me if there are no votes displayed, it just cuts it short with lots of space left. Not sure if we want to address it:
image

src/components/Proposal.vue
```diff
@@ -32,11 +32,11 @@ async function handleVoteClick(choice: Choice) {
               space: `${route.params.id}`
             }
           }"
-          class="max-w-fit flex space-x-2"
+          class="space-x-2 flex max-w-[700px]"
```
bonustrack, line 35:
```suggestion
          class="space-x-2 flex"
```

PR #822, opened 2023-11-29
feat: use Stamp API to resolve names

bonustrack:
Moving to Stamp RPC API to resolve names instead of doing requests on frontend, with this change we also resolve Lens and UnstoppableDomains names, and it includes a cache system. To test this change you can try delegates, votes page or others sections where we display ENS names.


Sekhmet (approved):
tACK

Issue #815, opened 2023-11-15
bug: can't favorite this space


bonustrack:
When I add in favorite this space, it does not appear on the sidebar even after reload: https://snapshotx.xyz/#/s:hydraventures.eth I can see the space within the localStorage.

PR #829, opened 2023-12-01
feat: use spaces query instead of ranking

Sekhmet:
### Summary

Ranking is better for explore page, but we don't use it on SX-UI. We can't filter items using ranking query, we have to use regular spaces query.

Closes: https://github.com/snapshot-labs/sx-ui/issues/815

### How to test

1. Favorite this space: http://localhost:8080/#/s:hydraventures.eth
2. It shows up on favorites page.


bonustrack (approved):
tACK

Issue #813, opened 2023-11-13
chore: remove Snapshot option for treasury or delegation network


bonustrack:
When creating a space there is "Snapshot" as option for treasury network or delegation, we shouldn't propose Snapshot here as it's not a source of treasury or delegation.

PR #830, opened 2023-12-01
fix: remove read-only networks from delegation/treasury enums

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/813

### How to test

1. Go to http://localhost:8080/#/create
2. You can't select Snapshot as network for treasury nor delegation network.


bonustrack (approved):
tACK

Issue #826, opened 2023-12-01
bug: cast vote with "abstain" is failing on SX-SN


bonustrack:
There seem to be an issue with "Abstain" choice, when I try to cast a vote it fail with an error from Mana "unauthorized".

PR #836, opened 2023-12-04
fix: convert choices to native format

Sekhmet:
### Summary

Now sx-starknet uses same enum as sx-evm so we need to convert it.

Closes: https://github.com/snapshot-labs/sx-ui/issues/826

### How to test

1. Create proposal and vote to abstain.
2. It works and you should see it as abstained vote.


bonustrack (approved):
tACK, works well now. If both EVM and SN versions use same enum can't we just use the proper choice number when someone cast a vote without converting it?

Sekhmet:
I think we should stop using numbers to encode choice entirely at UI level (including sx.js and sx-api) and instead use strings to avoid this entirely, but it will require a bit of refactoring on all projects. I guess we could start with sx.js to avoid those implicit conversions on UI entirely.

Issue #755, opened 2023-09-22
bug: clicking on SpaceItem doesn't open Space on first click on iOS Safari


Sekhmet:
There seem to be some issue on Safari as clicking on SpaceItem on explore page won't open space, just add hover state (to show favourite button), you then need to tap it for second time to actually open.

samuveth:
@bonustrack I can not reproduce this issue on my end.

MacOS Ventura Version 13.5.2 (22G91)
Safari Version 16.6 (18615.3.12.11.2)

samuveth:
I'd like to work on this but don't have an iOS device available.

Sekhmet:
Do you use macOS? If so I think you can test it with Safari and Xcode.

ChaituVR:
@Sekhmet you started working on this already? i can take this if not

Sekhmet:
@ChaituVR I did not, go a head. Thank you!

PR #837, opened 2023-12-05
fix: Disable hover on devices that don't support it

ChaituVR:
Closes https://github.com/snapshot-labs/sx-ui/issues/755

### Summary
Fix is related to https://github.com/tailwindlabs/tailwindcss/pull/8394
> Again this will be a breaking change in v4, but only for evil people who were writing code like hidden group-hover:block. Until then, everything will work the same, and you can opt-in to this if you want to start using it now.

So the problem is with the `group-hover` class and IOS devices don't support hover events, on the first click it assumes as `hover`

This change will disables on mobile devices but hovers should still work on desktops 
An alternate way is to handle click with JS

### How to test

1. Open the explore page on IOS Safari
2. Click on any space
3. Before fix: It will take two clicks to go to space page
4. After fix: It will take one click to go to space page
5. Also try same on desktops, it should display a star on hover and redirect to space page on click

### Self-review checklist
- [x] I have performed a full self-review of my changes
- [x] Tested on IOS safari
- [x] Tested on desktop chrome
- [x] Test on Android chrome
- [ ] I have tested my changes on a preview deployment
- [x] I have tested my changes on different screen sizes (sm, md)

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -12,6 +12,9 @@ const withOpacity = variableName => {
 };
 
 module.exports = {
+  future: {
+    hoverOnlyWhenSupported: true
+  },
   content: ['./index.html', './src/**/*.{js,ts,vue}'],
   darkMode: 'class',
   theme: {


Sekhmet:
> only for evil people who were writing code like hidden group-hover:block

image

Sekhmet (approved):
tACK

Issue #833, opened 2023-12-03
feat: Support multiple delegations


bonustrack:
On SX-UI in space creation and space profile settings we should allow controller to setup multiple delegations instead of just one. This is to address Starknet governance requirement which has delegation on 2 different networks. 

We should also add a param “Name” for each of the delegation. On the UI if there is more than 1 delegation setup we should add a tab system to switch between different delegation list.

PR #838, opened 2023-12-05
feat: index multiple delegations

Sekhmet:
### Summary

**Only Starknet was tested as of now.**

Closes: https://github.com/snapshot-labs/sx-ui/issues/833

Depends on https://github.com/snapshot-labs/sx-api/pull/205
Depends on https://github.com/snapshot-labs/sx-subgraph/pull/43

### How to test

1. Run local sx-api.
2. Run sx-ui with sn-tn only: `VITE_ENABLED_NETWORKS=sn-tn yarn dev`.
3. Deploy new space with multiple delegations.
4. Edit that space delegations in edit space modal.
5. View delegations for space (**currently subgraph fails for some subgraphs because it's overloaded**).

### Screenshots

image
image
image

### DIFF
--- a/src/components/Block/Delegates.vue
+++ b/src/components/Block/Delegates.vue
@@ -1,0 +1,182 @@
+<script setup lang="ts">
+import { _n, shorten } from '@/helpers/utils';
+import { getNetwork } from '@/networks';
+import { Space, SpaceMetadataDelegation } from '@/types';
+
+const props = defineProps<{ space: Space; delegation: SpaceMetadataDelegation }>();
+
+const delegateModalOpen = ref(false);
+const sortBy = ref(
+  'delegatedVotes-desc' as
+    | 'delegatedVotes-desc'
+    | 'delegatedVotes-asc'
+    | 'tokenHoldersRepresentedAmount-desc'
+    | 'tokenHoldersRepresentedAmount-asc'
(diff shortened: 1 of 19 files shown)


bonustrack:
Works well, some suggestions for the style:

- Use full rounded button for "Add new delegation API"
- Rename this button to "Add delegation" (on the modal title too)

Nice to have:

- Ideally we should not have a modal on top of a modal, this makes modal backdrop twice darker. I imagine if we figure out this it will be easier to reuse others modals like select token list modal without duplicate code.

Sekhmet:
> Ideally we should not have a modal on top of a modal, this makes modal backdrop twice darker. I imagine if we figure out this it will be easier to reuse others modals like select token list modal without duplicate code.

Do you want to just avoid second backdrop or make it more complex (like system where you open multiple modals at once, but only one is visible at the time etc.)? If it's latter I'd create separate issue for it.

bonustrack:
I mean the later

Sekhmet:
Looks like all subgraphs were updated other than arbitrum which is stuck (?).
I will hardcode it to specific version so all APIs have same format and this can be merged (because subgraphs have changed master deployment is now broken).

We should come back to see if it's unstuck later on and update it back to latest.

Issue #827, opened 2023-12-01
feat: support single slot proof strategy


bonustrack:
We need to add support for single slot proof voting strategy on the UI. This will requires some change in SX.js, Mana and SX-UI. 

Here is a script to run the whole flow with comments from @Orland0x: https://github.com/snapshot-labs/sx-starknet/blob/feat_herodotus_integration/tests/single-slot-proof.ts

The flow is working as following:

- Someone create a proposal on SX-UI
- The proposal start
- Mana send a request to Herodotus API with the proposal id and the proposal start timestamp to prove token storage.
https://github.com/snapshot-labs/sx-starknet/blob/ac7adbf5469f7722a079010704e23eb9fb6b30a2/scripts/herodotus-interaction-demo.ts#L78-L114
- Herodotus webhook server send a request to Mana when the query is processed 
https://github.com/snapshot-labs/sx-starknet/blob/ac7adbf5469f7722a079010704e23eb9fb6b30a2/scripts/herodotus-interaction-demo.ts#L106-L111
- Mana send a request to Herodotus API to get the binary search tree to remap the snapshot timestamp to the L1 block number 
https://github.com/snapshot-labs/sx-starknet/blob/ac7adbf5469f7722a079010704e23eb9fb6b30a2/scripts/herodotus-interaction-demo.ts#L131-L141
- Mana send a transaction to cache the block number in single slot proof voting strategy contract https://github.com/snapshot-labs/sx-starknet/blob/ac7adbf5469f7722a079010704e23eb9fb6b30a2/scripts/herodotus-interaction-demo.ts#L146-L166
- On SX-UI someone cast a vote, SX.js query the Ethereum node for getting storage proof data of the desired slot at the snapshot L1 block number 
https://github.com/snapshot-labs/sx-starknet/blob/ac7adbf5469f7722a079010704e23eb9fb6b30a2/scripts/herodotus-interaction-demo.ts#L168-L182

PR #841, opened 2023-12-11
feat: add evmSlotValue strategy support

Sekhmet:
### Summary

Depends on: https://github.com/snapshot-labs/sx.js/pull/243
Depends on: https://github.com/snapshot-labs/mana/pull/125
Depends on: https://github.com/snapshot-labs/sx-api/pull/206

This PR adds support for evmSlotValue strategy.

Closes: https://github.com/snapshot-labs/sx-ui/issues/827

### How to test

Set up all dependencies first and make sure you are using local mana and sx-api. Check mana/sx-api test plans for more details. There is lots of dependencies so it's a bit cumbersome to set up right now.

1. Create new space on Starknet (goerli) - mainnet is not supported by Herodotus using EVM Slot value strategy. You can use this token (`0xd96844c9b21cb6ccf2c236257c7fc703e43ba071` + slot index `0` as you can mint it yourself).
2. Create new proposal.
3. Check mana logs for internalId and webhook URL (will be automated on deployment).
4. Check `https://api.herodotus.cloud/batch-query-status?apiKey=API_KEY&batchQueryId=INTERNAL_ID_RETURNED_BY_MANA` until status is `DONE`.
5. Open webhook URL.
6. Refresh page, you should have VP.
7. You should be able to vote.


src/networks/starknet/constants.ts
```diff
@@ -393,7 +341,89 @@ export function createConstants(networkId: NetworkID) {
...
+                },
+                symbol: {
+                  type: 'string',
+                  maxLength: MAX_SYMBOL_LENGTH,
+                  title: 'Symbol',
+                  examples: ['e.g. COMP']
+                },
+                slotIndex: {
+                  type: 'integer',
+                  title: 'Slot Index',
```
bonustrack, line 413:
```suggestion
                  title: 'Slot index',
```

src/networks/starknet/constants.ts
```diff
@@ -393,7 +341,89 @@ export function createConstants(networkId: NetworkID) {
...
+            paramsDefinition: {
+              type: 'object',
+              title: 'Params',
+              additionalProperties: false,
+              required: ['contractAddress', 'slotIndex', 'chainId'],
+              properties: {
+                contractAddress: {
+                  type: 'string',
+                  format: 'address',
+                  title: 'Token address',
```
bonustrack, line 397: This can be used on a contract that is not a token, I would rather use this:
```suggestion
                  title: 'Contract address',
```

src/networks/starknet/constants.ts
```diff
@@ -393,7 +341,89 @@ export function createConstants(networkId: NetworkID) {
...
+                  maxLength: MAX_SYMBOL_LENGTH,
+                  title: 'Symbol',
+                  examples: ['e.g. COMP']
+                },
+                slotIndex: {
+                  type: 'integer',
+                  title: 'Slot Index',
+                  examples: ['0']
+                },
+                chainId: {
```
bonustrack, line 416: Can we force the chainId to be 5 on a Starknet testnet spaces and 1 on Starknet mainnet spaces and remove this param? Also I would recommend to put the params in this order: contractAddress, slotIndex, decimals, symbol

bonustrack (commented):
I'm not able to run the UI with SX.js linked, I got this error:
image

I can see the file `/starknet/actions` is firing type error with `NetworkActions`
image

bonustrack:
Can we remove EVM slot proof option in the proposal validation section? This is not something we can support.

bonustrack:
When I open the proposal page a request is failing to Herodotus API (`timestamp` and `accumulates_chain` are wrong):
```https://ds-indexer.api.herodotus.cloud/binsearch-path?timestamp=0&deployed_on_chain=SN_GOERLI&accumulates_chain=undefined```

When I cast a vote the request fail with this endpoint (`timestamp` is correct now but not `accumulates_chain`)
```https://ds-indexer.api.herodotus.cloud/binsearch-path?timestamp=1702408665&deployed_on_chain=SN_GOERLI&accumulates_chain=undefined```

I can see the message to sign doesn't include proof
image

When I hardcode `accumulates_chain=5` when I cast a vote the request to Herodotus API works but I still not see any proof or extra data in the message to sign, `UserVotingStrategies` is empty

When I confirm the vote, request to Mana fail with `unauthorized` error message with this logs:
```sh
 40: Contract error: {"revert_error":"Execution error at transaction index 0: Error in the called contract (0x06abd599ab530c5b3bc603111bdd20d77890db330402dc870fc9866f50ed6d2a):\nError at pc=0:15647:\nGot an exception while executing a hint.\nCairo traceback (most recent call last):\nUnknown location (pc=0:233)\nUnknown location (pc=0:5191)\nUnknown location (pc=0:11307)\n\nError in the called contract (0x048b33fe56e9b9354d4278ffdd5f6d546b13aa3d8c33149db2e2e2fdb48a369e):\nError at pc=0:2855:\nGot an exception while executing a hint: Custom Hint Error: Execution failed. Failure reason: 0x5573657220686173206e6f20766f74696e6720706f776572 ('User has no voting power').\nCairo traceback (most recent call last):\nUnknown location (pc=0:554)\nUnknown location (pc=0:1844)\n"}
```

bonustrack:
Also I can't see my voting power, it show 0 in proposals or proposal page.

bonustrack:
Is it possible to get the L1 block number for the contract directly (https://testnet.starkscan.co/contract/0x02ee57d848297bc7dfc8675111b9aa3bd3085e4038e475250770afe303b772af#read-write-contract) instead of calling Herodotus API?

Sekhmet:
> Is it possible to get the L1 block number for the contract directly (https://testnet.starkscan.co/contract/0x02ee57d848297bc7dfc8675111b9aa3bd3085e4038e475250770afe303b772af#read-write-contract) instead of calling Herodotus API?

We would need to pass tree to make this contract all which requires request to Herodotus (but this requests already gives us block number).

bonustrack:
Is the tree related to block number or you mean the storage proof?

Sekhmet:
Binary tree here is something Herodotus specific: https://docs.herodotus.dev/herodotus-docs/protocol-design/timestamp-to-block-number-mapper#binary-search

bonustrack:
To get current voting power on "proposals" page you can query storage from a L1 node with something like this: https://gist.github.com/bonustrack/bc5c770ba23b0d0026fbf2c9bd966bc8 with `latest` as `blockNum`

bonustrack:
I was able to cast a vote 🎉 
image
http://localhost:8080/#/sn-tn:0x039c7d023cfa2115b27b01d727820b6a2f13a7eadf04709cfb2a598fec41fe90/proposal/1/votes

bonustrack:
Voting power on proposal page working now. It just not visible on proposals page, this can be added later in another PR if you prefer.

bonustrack:
Found these 2 issue in the Mana logs, the 2nd trigger a restart of the server
image

bonustrack:
And voting power on proposals page works well now

Sekhmet:
Yup, debugging it right now.

Sekhmet:
Looks like webhook doesn't work like we expect it to work. It just forwards the response for the request we make (right away), not when request is DONE.

I guess we can disable webhooks for now on mana and merge this PR so we can interact with it (kind of manually) via mana logs, but we will need to handle that webhook manually.

bonustrack:
Sound good

bonustrack:
Looks like something wrong with sx-api too
image

bonustrack:
It's stuck in a loop
image

Sekhmet:
Fixed by re-deploy, we might need to increase delay between starting indexer as it seems that 1 minute is not enough.

Looks like Herodotus API is down.

Sekhmet:
Looks like API is back up.

bonustrack:
When I manually trigger Herodotus it restart Mana server here is the error:
image

bonustrack (approved):

PR #842, opened 2023-12-12
feat: show Voting ended label if voting has already ended

Sekhmet:
### Summary

Without extra label it's confusing what it means when it's just "1y ago". For current proposals it seems fine ("6h left").

### How to test

1. Go to http://localhost:8080/#/arb1:0x2fb231a3ac7c40888d2e6773f5600851b0c20ce4/proposal/2 (numbers don't make sense, ignore that :D)

### Screenshots

image

### DIFF
--- a/src/views/Proposal/Overview.vue
+++ b/src/views/Proposal/Overview.vue
@@ -50,6 +50,19 @@ const proposalMetadataUrl = computed(() => {
   return sanitizeUrl(url);
 });
 
+const votingTime = computed(() => {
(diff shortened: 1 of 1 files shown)


pscott:
`Voting ended 1y ago` or `Ended 1y ago`?
While I do agree being explicit is often better, I'm not sure about this particular one, wdyt?

Sekhmet:
I think I prefer `Ended` as well.

ChaituVR (approved):
tAck

Issue #386, opened 2023-01-24
feat: populate execution with WalletConnect


bonustrack:
**This issue is made for hiring challenge only, if you do this challenge make the change in a forked repository, don't PR this repository.**

We should add a feature to be able to login as a DAO to dapps like Uniswap, with this account and network (Goerli): https://github.com/snapshot-labs/sx-ui/blob/master/src/helpers/space.json

- On the proposal creation page we should add a button "WalletConnect" in the execution section with a `login` icon see:
<img width="1439" alt="image" src="https://user-images.githubusercontent.com/16245250/214299898-896af7ba-8dca-4dc6-a17c-92dd0386ff80.png">

- When we click on this button it should open a modal with a form to input a connection link like this:
<img width="503" alt="image" src="https://user-images.githubusercontent.com/16245250/214301331-84e7b06d-e512-4fd5-955e-8e5067ec7257.png">

The connection link is used to connect SX UI to a dapp, you can get a connection link from any dapp supporting WalletConnect like Uniswap (on Goerli) here:
<img width="614" alt="image" src="https://user-images.githubusercontent.com/16245250/214303430-3551845c-bfbe-4873-906f-42e61df11f5e.png">

- Once the user input a connection link and click confirm, we need to use WalletConnect SDK to log as the DAO into Uniswap. Here is an example on how to do that: https://github.com/snapshot-labs/guest/blob/master/src/composables/useWalletConnect.ts#L83-L86 This come from http://guest.so which is a simple app to be able to login with any account on dapps.
We can add a loading icon into the "Confirm" button while login is going on.

- Once login is done we should close the modal
- If the modal is open again, we should show the logged account (the DAO) and a logout button like this:
<img width="503" alt="image" src="https://user-images.githubusercontent.com/16245250/214305603-6799f0e0-8d81-4d12-8852-275de56877ff.png">

- If we click logout, it should logout the DAO to the dapp and come back to the previous form with Connection link input.

- If we trigger a new transaction on the dapp it should populate the execution section with new transaction (you can test that by trying to do a swap on Uniswap for example), it would look like this:

<img width="648" alt="image" src="https://user-images.githubusercontent.com/16245250/214306103-7f686ac0-baf3-42ed-993d-cc8ef800b3a7.png">

[...]

PR #854, opened 2024-01-16
feat: add WalletConnect execution

Sekhmet:
### Summary

Closes: https://github.com/snapshot-labs/sx-ui/issues/386

Looking for approach/design approvals foremost.

### How to test

1. Go to treasury page of this space: http://localhost:8080/#/sep:0xeb8ea538c90d2e009460ed69add7025560e18481/treasury
2. Go to Uniswap on speolia: https://app.uniswap.org/swap?chain=sepolia
3. Connect both, create some execution.

### Screenshots

image
image
image
image
image


bonustrack:
- When I click approve, I can't see any loading within the approve button, it appear full white
image
- I was not able to login on Uniswap Sepolia or Aave (https://staging.aave.com/markets/) for some reason it get stuck when I click approve, I see this error in the console:
image

`index-b727bc3d.js:573 Uncaught (in promise) Error: Non conforming namespaces. approve() namespaces chains don't satisfy required namespaces.
      Required: eip155:1
      Approved: eip155:11155111`
      
- For the button icon we should use WalletConnect logo directly, here is the official SVG: https://github.com/WalletConnect/walletconnect-assets/blob/master/Logo/Blue%20(Default)/Logo.svg ideally the icon would use link color (white and black)

- We can change the tooltip and modal title to "Connect to apps"
- For the URL it looks soso with mono I would instead make it simple like this:
image

Sekhmet:
Should work now, I wasn't able to test with Aave because WalletConnect doesn't open on that URL.

bonustrack:
It work for me now on Uniswap but only to swap ETH to UNI, when I try to wrap ETH to WETH it doesnt trigger a new transaction.

For the flow what would be better is to keep the WalletConnect connection alive regardless of the page you staying on SX, currently if I change page it will break connection. Also when I trigger a new tx on Uniswap (or other app) we should show a modal with "Add transaction" populated with the tx and confirm button, and this even if I'm still on the treasury page. We should redirect user to proposal creation page only if user confirm the new tx.

Otherwise modal title should be "Connect to apps"

Sekhmet:
> For the flow what would be better is to keep the WalletConnect connection alive regardless of the page you staying on SX, currently if I change page it will break connection.

I'm not sure how to handle that UX-wise. Connection needs to be per-space (because we are connecting to app as treasury from specific space). So what happens if we navigate from the space? And if we navigate from it how do we disconnect? If we put the button on topnav it might end up being confusing (as it doesn't look space-specific anymore).

Do you think any specific idea in mind?

bonustrack:
If we move to another space then we wont see the connection in WC modal, another session can be started from that space. But if you confirm a transaction you get redirected to the editor for the correct space

bonustrack:
Looks good! Some others improvements:

- We should have the ability to add more than 1 tx, currently adding another one delete previous one.
- Instead of showing the modal with "New transaction has been detected" we should show the modal "Add transaction" directly. And if user click "Confirm" it get added in the proposal.
- Uniswap app icon doesn't load for me, for some reason it was working previously. I can see Metamask return this image: https://app.uniswap.org/favicon.png and not a .ico image.

image

Sekhmet:
@bonustrack Uniswap updated their metadata on WalletConnect and their image they return first no longer resolves.

bonustrack:
Ok then no worry will probably be fixed at some point

bonustrack:
Somehow something is wrong with the background, showing white color on scroll see: https://www.loom.com/share/8a6dbcd305c84c108ce6cf857af58fe7

I'm now able to add multiple txs, but it's hard to review what's the tx is about, what I recommend is to show this modal directly when a new tx is triggered: 
image

Sekhmet:
> Somehow something is wrong with the background, showing white color on scroll see: https://www.loom.com/share/8a6dbcd305c84c108ce6cf857af58fe7

This was fixed there: https://github.com/snapshot-labs/sx-ui/pull/858

bonustrack (approved):
tACK

PR #858, opened 2024-01-29
fix: apply dark class to html tag

Sekhmet:
### Summary

Since https://github.com/snapshot-labs/sx-ui/pull/794 dark style is only applied within Vue tree, which won't apply to body itself.

This causes body to remain white when dark theme is used which can become visible during overscroll. This commit replaces two ways of setting dark theme (data-color-scheme and class) with just once (class) set to html document.

### How to test

1. Go to UI, can switch between dark/light mode with no issues.
2. Run `yarn story:dev` - dark mode works there too.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -4,7 +4,6 @@ const el = ref(null);
 const route = useRoute();
 const uiStore = useUiStore();
 const { modalOpen } = useModal();
-const { userSkin } = useUserSkin();
 const { init, app } = useApp();
 const { web3, web3Account } = useWeb3();
 const { loadVotes, votes } = useAccount();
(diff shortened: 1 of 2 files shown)


samuveth:
I'm not able to test if this fixes the issues, as I said on discord
> I'm not able to reproduce this scrolling, for me it doesn't scroll at all on chrome macOS

Also I was wondering is there any reason the `window.matchMedia` is commented out in useSkin? Wouldn't it be easier to simply use `useDark` from vueUse https://vueuse.org/core/useDark/ which combines [usePreferredDark](https://vueuse.org/core/usePreferredDark/) and [useStorage](https://vueuse.org/core/useStorage/)

Sekhmet:
I think overscroll only happens when using trackpad and it's pretty common default behaviour on macOS.

`useDark` would be actually nice if we end up using `matchMedia` - we removed matchMedia 2 years ago, not sure why @bonustrack

samuveth (approved):
tACK

PR #861, opened 2024-02-01
refactor: replace choice magic numbers with string values

Sekhmet:
### Summary

We used to use 1 | 2 | 3 for vote choice on UI (for, against, abstain). Which was inconsistent with sx.js (0 abstain, 1, for, 2 against).

It could lead to issues as those values are not really clear on the UI. This commit replaces 1 | 2 | 3 with 'for' | 'against' | 'abstain' that are converted to SDK values in single place. This makes UI code more readable as actions are easy to understand.

### How to test
- You can still vote -> signed message (if using sig authenticator) uses correct choice.


bonustrack (approved):
tACK

PR #862, opened 2024-02-02
refactor: use defineModel

Sekhmet:
### Summary

This is now stable and recommended method of using model in components.

https://vuejs.org/guide/components/v-model.html

Upgraded deps to match new Vue version.

### How to test

1. Everything works as it used to, probably the best flow is to test creating a space and creating a proposal.


samuveth (approved):
Nice!
utACK

bonustrack (approved):
tACK

PR #863, opened 2024-02-02
test: colocate test files next to source files

Sekhmet:
### Summary

With mostly flat structure I feel like colocating those files makes the most sense, especially with also supporting story files. This way all files are close to each other:
- Button.ts
- Button.story.ts

or
- utils.ts
- utils.test.ts


samuveth (approved):
👍 
utACK

PR #866, opened 2024-02-05
fix: Hero light mode

samwhosung:
### Summary

The hero on the Homepage was not working on light mode, might have been caused by switching to the `.dark` class for the css variables in https://github.com/snapshot-labs/sx-ui/commit/cbcc39fe6446f23ef5eb09490995af59e8dcbeb6

### DIFF
--- a/src/style.scss
+++ b/src/style.scss
@@ -173,19 +173,6 @@ a,
   @apply bg-skin-text animate-pulse-fast;
 }
 
-.hero {
-  background-repeat: repeat;
-  background-position: top center;
-}
-
-.dark .hero {
-  background-image: url('@/assets/grid-dark.svg');
-}
-
-.light .hero {
-  background-image: url('@/assets/grid-light.svg');
-}
-
 .no-scrollbar::-webkit-scrollbar {
   display: none;
 }
--- a/src/style.scss
+++ b/src/style.scss
@@ -173,6 +173,16 @@ a,
   @apply bg-skin-text animate-pulse-fast;
 }
 
+.hero {
+  background-repeat: repeat;
+  background-position: top center;
+  background-image: url('@/assets/grid-light.svg');
+}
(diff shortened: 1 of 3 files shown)


Sekhmet (changes requested):
This was regression in https://github.com/snapshot-labs/sx-ui/pull/858 as since this PR there is no longer `light` class (just default style + additional `dark` class. There are no other usages of light, so I think it makes sense to default to light styles and on-demand update styles when `dark` is set (that's what we do with `:root` styles and `dark` class.

I think we could just fix it by this change (on top of master):
```diff
diff --git a/src/style.scss b/src/style.scss
index 78a5824..516bd1f 100644
--- a/src/style.scss
+++ b/src/style.scss
@@ -176,16 +176,13 @@ a,
 .hero {
   background-repeat: repeat;
   background-position: top center;
+  background-image: url('@/assets/grid-light.svg');
 }
 
 .dark .hero {
   background-image: url('@/assets/grid-dark.svg');
 }
 
-.light .hero {
-  background-image: url('@/assets/grid-light.svg');
-}
-
 .no-scrollbar::-webkit-scrollbar {
   display: none;
 }
```

And no longer use `light` class.

samuveth:
Yeah I see. Any reason to change it back otherwise I would leave it how I did now

Sekhmet:
@samuveth Personally I'd revert it, because it's spreads dark handling into more and more places (it's now being handled by `dark` class (all we should need), but this adds `dark:` prefix, extra classes in tailscale config just for background - a bit too much).

`dark` class (and CSS variables controlled by it) should be primary mean of adjusting styles (`dark:` usage in Alert should be migrated at some point too if we have design system variables for those).

samuveth:
Done

Sekhmet (approved):
tACK

PR #870, opened 2024-02-05
fix: Keyboard navigation for buttons

samwhosung:
### Summary

Fixes keyboard navigation for button, before outline was disabled everywhere and now it's just disabled for inputs. Also added some buttons where I found them to be missing.

### DIFF
--- a/src/components/Block/SpaceFormNetwork.vue
+++ b/src/components/Block/SpaceFormNetwork.vue
@@ -19,16 +19,17 @@ const availableNetworks = enabledNetworks
   <div class="mb-2">
     <h3>Space network</h3>
     <div class="grid grid-cols-auto gap-3 pt-4 mb-3">
-      <div
+      <button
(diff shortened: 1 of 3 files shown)


Sekhmet (commented):
With this PR combined https://github.com/snapshot-labs/sx-ui/pull/867 it won't have hover state for buttons (UiButtons).

Also this doesn't look that good now (just focusing by mouse, not tabbed in):
image

samuveth:
There was no hover state before either :) Do you want me to add one in this PR?

samuveth:
Oh I see, would you prefer we remove the `outline-0` from button or use `focus-visible:border-skin-link` which will only focus when using keyboard?

It would behave consistently with other buttons in the UI if we just remove the `outline-0` so I would go with that.
image

samuveth:
I moved these changed into https://github.com/snapshot-labs/sx-ui/pull/867 for easier testing

